PR #10, opened 2020-02-26
Validate cleaned CSDL metadata with typewriter

MIchaelMainer:
* Created v1.0 validation pipeline - this has run successfully.
* Created beta validation pipeline - this still needs validation, requires XSLT changes.
* Created script to download and run the latest release of typewriter.
* Created script to copy generated files.

Script example usage in script files. Expect that we will need to still
validate these changes against a scheduled build to make sure it is
configured appropriately.

	new file:   pipelines/validateCleanedMetadata_beta.yml
	new file:   pipelines/validateCleanedMetadata_v1.0.yml
	new file:   scripts/copyGeneratedFiles.ps1
	new file:   scripts/runTypewriter.ps1

### DIFF
--- a/scripts/copyGeneratedFiles.ps1
+++ b/scripts/copyGeneratedFiles.ps1
@@ -4,7 +4,7 @@
 <#
 .Synopsis
     Moves files from a source array of directories into the destination array
-    of directories.
+    of directories.  NOTE: all files will be removed from the destination array prior to copy.
 
 .Example
     .\scripts\copyGeneratedFiles.ps1 @("D:\temp\democopy\from\1", "D:\temp\democopy\from\2") @("D:\temp\democopy\to\3", "D:\temp\democopy\to\4")
--- a/scripts/copyGeneratedFiles.ps1
+++ b/scripts/copyGeneratedFiles.ps1
@@ -1,0 +1,58 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+<#
+.Synopsis
+    Moves files from a source array of directories into the destination array
+    of directories.
+
+.Example
+    .\scripts\copyGeneratedFiles.ps1 @("D:\temp\democopy\from\1", "D:\temp\democopy\from\2") @("D:\temp\democopy\to\3", "D:\temp\democopy\to\4")
+
+.Parameter fromPath
+    Required. An array of strings that represent the fully qualified directory
+    paths for copying generated files. The order of directory names is significant.
+
+.Parameter toPath
+    Required. An array of strings that represent the fully qualified destination
+    directory paths for copying generated files. The order of directory names is significant.
+#>
+Param(
+    [parameter(Mandatory = $true)][string[]]$fromPath,
+    [parameter(Mandatory = $true)][string[]]$toPath
+)
+
+# Validate arguments
+if ($fromPath.Count -ne $toPath.Count)
+{
+    Write-Host "The fromPath and toPath arrays must have matching elements." -ForegroundColor Red
+    Write-Host "##vso[task.complete result=Failed;]DONE"
+    exit 1
+}
+$allPaths = $fromPath + $toPath
+foreach ($dirPath in $allPaths)
+{
+    if ((Test-Path -PathType Container -Path $dirPath) -ne $true)
+    {
+        Write-Host "Argument contains invalid path to a directory: $dirPath" -ForegroundColor Red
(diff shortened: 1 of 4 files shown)


scripts/copyGeneratedFiles.ps1
```diff
@@ -0,0 +1,58 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+<#
+.Synopsis
+    Moves files from a source array of directories into the destination array
+    of directories.
```
ddyett:
```suggestion
    of directories.  NOTE: all files will be removed from the destination array prior to copy.
```

scripts/runTypewriter.ps1
```diff
@@ -0,0 +1,150 @@
...
+    See https://github.com/microsoftgraph/MSGraph-SDK-Code-Generator#using-typewriter for more information.
+
+.Parameter output
+    Optional. The output path of the generated files. Typewriter creates this directory if it doesn't exist.
+    See https://github.com/microsoftgraph/MSGraph-SDK-Code-Generator#using-typewriter for more information.
+
+.Parameter generationMode
+    Optional. Specifies the typewriter.exe generation mode.
+    See https://github.com/microsoftgraph/MSGraph-SDK-Code-Generator#using-typewriter for more information.
+#>
```
ddyett: missing one last description for the cleanup parameter.

scripts/runTypewriter.ps1
```diff
@@ -0,0 +1,150 @@
...
+$gh_owner_and_repo = 'microsoftgraph/MSGraph-SDK-Code-Generator'    # Org and repo where typewriter.exe is released.
+$tempDir           = '.\temp'
+$LASTEXITCODE      = $null
+$typewriterZipDrop = "$tempDir\typewriter.zip"
+$typewriterFilesDir = "$tempDir\Release\*"
+$typewriter = Join-Path $tempDir "typewriter.exe"
+
+# Create our temporary working directory.
+if ((Test-Path -PathType Container -Path $tempDir)) {
+    # Remove-Item $tempDir -Force -Recurse
```
ddyett: should this be enabled so that when we leave this if/else block we are in exactly the same state regardless of creation or previous existence.

scripts/runTypewriter.ps1
```diff
@@ -0,0 +1,150 @@
...
+    [Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12
+    $jsonObject = Invoke-WebRequest -Uri $feedQuery -UseBasicParsing | ConvertFrom-Json
+
+    # Download typewriter from the latest GitHub release.
+    if ($jsonObject.assets[0].name -eq 'typewriter.zip') {
+        $downloadURL = $jsonObject.assets[0].browser_download_url # GitHub release API provides the latest
+        Invoke-WebRequest -Uri $downloadURL -OutFile $typewriterZipDrop -UseBasicParsing -Verbose | Write-Host
+    }
+    else {
+        Write-Error 'typewriter.zip was not found using the release API. Check the release on GitHub. Make sure that
```
ddyett: shouldn't we terminate if this happens instead of continuing on to expand the archive.  I can't recall if Write-Error actually terminates so I could just be missing that.

scripts/runTypewriter.ps1
MIchaelMainer: That's how I originally had it set up, but then realized that if I want to run typewriter multiple times in a pipeline, I don't necessarily want to download and unzip again every time I run this script.

scripts/runTypewriter.ps1
MIchaelMainer: Yes! Thank you

scripts/runTypewriter.ps1
ddyett: that makes sense but you are running the risk of a dirty state causing an issue.  In a non-test environment cleaning it up may work out be more stable and not as horrible to keep purging.  Technically with the clean up flag being true you already do this.

---
In reply to: [385447267](https://github.com/microsoftgraph/msgraph-metadata/pull/10#discussion_r385447267) [](ancestors = 385447267)

MIchaelMainer: Even with the clean up flag, we can be dirty. The case where this could be a problem is if there is a fix in typewriter and typewriter isn't cleaned up, we may not download the fixed typewriter.
Pushed a fix

PR #11, opened 2020-02-27
Remove namespaces, use aliased microsoft.graph namespace

MIchaelMainer:


transforms/csdl/preprocess_csdl.xsl
```diff
@@ -97,12 +101,12 @@
 
     <!-- These actions have the same parameters that need reordering. Will need to create a new template
          for each reordering. -->
-    <xsl:template match="edm:Action[@Name='accept'][.//edm:Parameter[@Name='bindingParameter'][@Type='microsoft.graph.event']]|
-                         edm:Action[@Name='decline'][.//edm:Parameter[@Name='bindingParameter'][@Type='microsoft.graph.event']]|
-                         edm:Action[@Name='tentativelyAccept'][.//edm:Parameter[@Name='bindingParameter'][@Type='microsoft.graph.event']]">
+    <xsl:template match="edm:Action[@Name='accept'][.//edm:Parameter[@Name='bindingParameter'][@Type='graph.event']]|
```
ddyett: >graph.event [](start = 102, length = 11)

i think your test sample is using the microsoft.graph.event so you should update that.

PR #23, opened 2020-06-18
Do not remove namespaces in XSL tranformation

zengin:
- Dropped the transformation that removes namespaces
- Added a script to run XSL transformation from command line
- Updated readme to refer to the new script

[AB#5024](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/5024)

### DIFF
--- a/transforms/csdl/preprocess_csdl.xsl
+++ b/transforms/csdl/preprocess_csdl.xsl
@@ -80,21 +80,17 @@
     </xsl:template>
 
     <!-- Remove NavigationProperty-->
-      
+
     <xsl:template match="edm:Schema[@Namespace='microsoft.graph']/edm:EntityType[@Name='approval']/edm:NavigationProperty[@Name='request']"/>
-  
+
     <!-- Remove all capability annotations-->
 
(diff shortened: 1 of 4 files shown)


zengin:
@MIchaelMainer, my change in the XSL doesn't explain why there is a diff for the `<EntityContainer>` element. I don't see a rule which removes that. Is it just that the output file was stale?

zengin:
@ddyett Azure Boards bot doesn't seem to be updating the task link for this PR. Do we miss the link between the repo and the board?

MIchaelMainer:
That output file was stale. The EntityContainer should've been there. Your checkin corrects it. 

The transform via script is way better!

zengin:
Not merging this until downstream handling of multiple namespaces are ready in Typewriter.

MIchaelMainer:
@zengin I want your script and I'm impatient so I'm going to add it in another PR.

zengin:
I am closing this pull request.

I will create another one with 2 transforms, one that drops namespaces, another that doesn't. After reaching to a state where we have steady releases, we will go back to a single transform that keeps all the namespaces.

PR #24, opened 2020-07-08
Add transform script, remove property, add action

MIchaelMainer:
The createUploadSession action was set with an additional parameter which is a breaking change for Java since Java doesn't support optional parameters (which only works when there aren't any non-nullable parameters after a nullable parameter).  I've added the original overload and have requested from the API owner that they add it back into the metadata.

Also removing an erroneous property from changeNotification, which should have been removed this week.

Added script from https://github.com/microsoftgraph/msgraph-metadata/pull/23.

@nikchett


MIchaelMainer:
Just set protected branch for master, and dismissing review if there are new changes. Need new review.

zengin:
@MIchaelMainer, does the C# Beta SDK from this metadata transform build for you? I tried it to make it a comparison base for my work, and I am getting a compilation failure.

MIchaelMainer:
@zengin I wouldn't be surprised if it didn't as the metadata was updated since this issue first arose. Let me try again with latest and then I will report back.

MIchaelMainer:
@zengin One new metadata, and a typo in transform. Making the fix and verifying now.

transforms/csdl/preprocess_csdl.xsl
```diff
@@ -97,6 +97,7 @@
     <!-- Remove namespaces -->
 
     <xsl:template match="edm:Schema[@Namespace='microsoft.graph.callRecords']"/>
+    <xsl:template match="edm:Schema[@Namespace='microsoft.graph.termStore']"/>
```
zengin, line 100: We are getting a new namespace. Great! Real life testing data for me :)

PR #30, opened 2020-09-18
bugfix/exclude-site-delta-overload

baywet:
realted to https://github.com/microsoftgraph/msgraph-beta-sdk-java/pull/20


baywet:
@zengin @MIchaelMainer FYI I originally had run the XSLT on the raw/clean output but I think it's a multi-step process and on top of removing the function I specifically wanted to remove, it removed a bunch of annotations for description (Lines 48000 and ongoing). I consequently rolled back that change and I guess we'll do it by triggering the CI. Let me know if I'm missing anything.

zengin:
@baywet yeah, the full CI pipeline has a TransformWithDocs step and that is probably what is missing in your local run steps. 
We have a test input-output pair in the CSDL folder. Do you mind adding your case into the test XML? There is a PowerShell script in the same folder to quickly run the transform as well.

Note to self: It would be worth thinking about writing a PowerShell script to go end to end and generate a quick diff compared to the last transform (may need snapshotting docs repo to avoid diffs in descriptions).

baywet:
Added, merging. Thanks for the review! :)

PR #32, opened 2020-09-24
bugfix/message-has-stream

baywet:
fixes https://github.com/microsoftgraph/msgraph-sdk-java/issues/323
related https://github.com/microsoftgraph/msgraph-sdk-java/pull/499

### DIFF
--- a/transforms/csdl/preprocess_csdl.xsl
+++ b/transforms/csdl/preprocess_csdl.xsl
@@ -113,6 +113,14 @@
       </xsl:copy>
     </xsl:template>
 
+    <!-- Add  attribute -->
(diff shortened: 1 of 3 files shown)


zengin (approved):
This is going to fix compilation of couple of snippets as well: https://github.com/microsoftgraph/msgraph-sdk-raptor/blob/dev/TestsCommon/TestDataGenerator.cs#L232

baywet:
@zengin should I PR there as well or will you do it?

zengin:
@baywet I will be notified by the test results once this goes into effect. I will clean up after that. Thanks for checking though.

PR #33, opened 2020-09-30
feature/odata-cast-type-constraint

baywet:
related https://github.com/microsoftgraph/msgraph-sdk-java/pull/509

### DIFF
--- a/transforms/csdl/preprocess_csdl.xsl
+++ b/transforms/csdl/preprocess_csdl.xsl
@@ -171,7 +171,6 @@
           <Collection>
             <String>microsoft.graph.appRoleAssignment</String>
             <String>microsoft.graph.application</String>
-            <String>microsoft.graph.directoryObjectPartnerReference</String>
             <String>microsoft.graph.endpoint</String>
             <String>microsoft.graph.group</String>
             <String>microsoft.graph.servicePrincipal</String>
--- a/transforms/csdl/preprocess_csdl.xsl
+++ b/transforms/csdl/preprocess_csdl.xsl
@@ -121,6 +121,19 @@
       </xsl:copy>
     </xsl:template>
 
+    <!-- Add odata cast annotation -->
+    <xsl:template match="edm:Schema[@Namespace='microsoft.graph']/edm:EntityType[@Name='group']/edm:NavigationProperty[@Name='members']">
+      <xsl:copy>
+        <xsl:copy-of select="@* | node()" />
+        <Annotation Term="Org.OData.Validation.V1.DerivedTypeConstraint">
+          <Collection>
+            <String>microsoft.graph.user</String>
+            <String>microsoft.graph.group</String>
+          </Collection>
+        </Annotation>
+      </xsl:copy>
+    </xsl:template>
+
     <!-- Remove attribute -->
     <xsl:template match="edm:Schema[@Namespace='microsoft.graph']/edm:EntityType[@Name='onenotePage']/@HasStream|
                          edm:Schema[@Namespace='microsoft.graph']/edm:EntityType[@Name='onenoteResource']/@HasStream">
--- a/transforms/csdl/preprocess_csdl.xsl
+++ b/transforms/csdl/preprocess_csdl.xsl
@@ -122,13 +122,145 @@
     </xsl:template>
 
     <!-- Add odata cast annotation -->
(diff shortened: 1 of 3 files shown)


zengin:
1. Are the annotations going to be managed by the XSLT going forward or is there a plan to integrate this into the original metadata? If it is XSLT, what is the expected load on our team to manage it here?
2. What is the source of truth for these annotations? I am partially aware of how it is defined, but it would be great if we call out in the PR the rule that determines which entities get these annotations.

zengin:
@baywet so this will be a subset of the exhaustive list of annotations that are expected to appear after engaging all the service workloads. It will unblock some of the common scenarios (at least the ones appearing in docs I presume), but we might be missing some of them and they can be added later on here. Depending on where the conversations land, this logic may move upstream to the actual metadata. Is that a fair understanding?

Can we call the format for the annotations final? In other words, how much tweaking will we need in Vipr/generator, if there are any suggestions on the format as part of the aforementioned conversations?

I am trying to understand the trade-off between taking this as a proof of concept implementation and unblocking known scenarios vs designing this end-to-end and doing it once. Any insights into this trade-off?

baywet:
> @baywet so this will be a subset of the exhaustive list of annotations that are expected to appear after engaging all the service workloads. It will unblock some of the common scenarios (at least the ones appearing in docs I presume), but we might be missing some of them and they can be added later on here. Depending on where the conversations land, this logic may move upstream to the actual metadata. Is that a fair understanding?

yes, that's exactly the intent.

> Can we call the format for the annotations final? In other words, how much tweaking will we need in Vipr/generator, if there are any suggestions on the format as part of the aforementioned conversations?

Sorry, I'm not sure I understand the question, can you rephrase? If the question is "is this annotation specified in OData?" It looks like it it, and it's implemented in the Odata lib as I'm using it.

> I am trying to understand the trade-off between taking this as a proof of concept implementation and unblocking known scenarios vs designing this end-to-end and doing it once. Any insights into this trade-off?

I don't see that as a proof of concept as I didn't come up with the annotation myself, see https://github.com/oasis-tcs/odata-vocabularies/blob/master/vocabularies/Org.OData.Validation.V1.xml#L147

MIchaelMainer:
DerivedTypeConstraint is well defined: https://github.com/oasis-tcs/odata-vocabularies/blob/7b876e2a97e13a7dd9be458cec8d1e204eacc35f/vocabularies/Org.OData.Validation.V1.md. I think we can call the annotation format final.

I think we are safe taking a dependency on this and incorporating it end to end.

I would like for us to start segregating XSLT scenarios:
preprocess_csdl.xsl - top level XSLT that is called from transform processes. Imports all other XSLT.
-- metadataFix.xsl - where we inventory and add metadata to fix issues in the production metadata/
-- toolingFix.xsl - where we inventory and alter metadata that works around issues in our generation tools.
-- apiWorkaround.xsl - where we inventory and alter metadata that works around OData non-conformance.

baywet:
@MIchaelMainer thanks for jumping in!
So in your suggestion, where would my changes go ? (which file)

zengin:
@baywet The fact that this is OData standard annotation answers second part of my question. Thanks for clarifying that.

I think the PR description part is a good place to connect the dots between the problem description and how the PR is attempting to resolve it. Something along the lines of:
- Here is an issue (a Github issue link is fine if it is descriptive enough)
- Here is how this PR is attempting to resolve it using such and such. This "such and such" could be a language or tooling feature, a new algorithm, an approach from a blog post etc.

Over communication in the second part is especially useful not only for easing into the reviews but also for keeping the history in a one stop shop, that is, when others look at the history going through the PRs will have context immediately from the PR description. My two cents.

baywet:
@zengin I plead guilty of being lazy sometimes and not providing enough context in the PR description 🙌

Most of the context was provided in https://github.com/microsoftgraph/MSGraph-SDK-Code-Generator/pull/303 even if I agree with the fact it probably wasn't enough.

Let me know if I need to change anything in this PR or whether it's good to merge and thanks for asking all those questions!

MIchaelMainer:
@baywet Let's put this into an imported metadataFix.xsl

baywet:
@MIchaelMainer do you mean in this PR or in a subsequent one?

MIchaelMainer:
@baywet Let's do it in a subsequent one. I'm not sure if those are all the scenarios we need to cover and we might as well make the change all at once. We need to review the different scenarios.

MIchaelMainer (approved):

zengin (approved):

PR #39, opened 2020-10-19
Consistency level header

### LINKED ISSUE #38: Add ConsistencyLevel header for advanced queries
In order for SDKs such as PowerShell to support advanced queries, the CSDL needs to have a [custom header annotation](https://github.com/oasis-tcs/odata-vocabularies/blob/master/vocabularies/Org.OData.Capabilities.V1.xml#L938) nested under a read restrictions on all entitySets that accept a consistencyLevel header. 

AGS does not currently expose all readRestrictions annotations in the CSDL. As a workaround, we should add customHeader to the following entitySets via XSLT:
- applications
- administrativeUnits
- contacts
- devices
- groups
- servicePrincipals
- users

``` xml
 <EntitySet Name="users" EntityType="Microsoft.DirectoryServices.user">
              <Annotation Term="Org.OData.Capabilities.V1.ReadRestrictions">
                <Record>
                  <PropertyValue Property="CustomHeaders">
                    <Collection>
                      <Record>
                        <PropertyValue Property="Name" String="ConsistencyLevel" />
                        <PropertyValue Property="Description" String="Indicates the requested consistency level." />
                        <PropertyValue Property="DocumentationURL" String="https://docs.microsoft.com/en-us/graph/api/orgcontact-list?view=graph-rest-beta" />
                        <PropertyValue Property="Required" Bool="false" />
                        <PropertyValue Property="ExampleValues">
                          <Collection>
                            <Record>
                              <PropertyValue Property="Value" String="eventual" />
                              <PropertyValue Property="Description" String="$search and $count queries require the client to set the ConsistencyLevel HTTP header to 'eventual'." />
                            </Record>
                          </Collection>
                        </PropertyValue>
                      </Record>
                    </Collection>
                  </PropertyValue>
                </Record>
              </Annotation>
        </EntitySet>
```

 https://developer.microsoft.com/en-us/microsoft-365/blogs/build-advanced-queries-with-count-filter-search-and-orderby/ 
[AB#6355](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/6355)

peombwa:
This PR adds consistency level header to the following AAD entity sets:
- applications
- administrativeUnits
- contacts
- devices
- groups
- servicePrincipals
- users

Closes #38

### DIFF
--- a/transforms/csdl/preprocess_csdl.xsl
+++ b/transforms/csdl/preprocess_csdl.xsl
@@ -357,6 +357,50 @@
       </xsl:element>
     </xsl:template>
 
+    <!-- Add custom headers (ConsistencyLevel) to AAD objects -->
+    <xsl:template name="ConsistencyLevelHeaderTemplate">
+      <xsl:element name="Record" namespace="{namespace-uri()}">
+        <xsl:element name="PropertyValue">
+          <xsl:attribute name="Property">CustomHeaders</xsl:attribute>
+          <xsl:element name="Collection">
+            <xsl:element name="Record">
+              <xsl:element name="PropertyValue">
+                <xsl:attribute name="Property">Name</xsl:attribute>
+                <xsl:attribute name="String">ConsistencyLevel</xsl:attribute>
+              </xsl:element>
+              <xsl:element name="PropertyValue">
+                <xsl:attribute name="Property">Description</xsl:attribute>
+                <xsl:attribute name="String">Indicates the requested consistency level.</xsl:attribute>
+              </xsl:element>
+              <xsl:element name="PropertyValue">
+                <xsl:attribute name="Property">DocumentationURL</xsl:attribute>
+                <xsl:attribute name="String">https://developer.microsoft.com/en-us/office/blogs/microsoft-graph-advanced-queries-for-directory-objects-are-now-generally-available/</xsl:attribute>
+              </xsl:element>
+              <xsl:element name="PropertyValue">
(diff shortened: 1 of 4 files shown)


zengin (approved):
I really wish this database replication concept (eventual consistency) didn't go all the way to the SDK surface area, but this PR makes it one step closer to hiding that information from customers.

@darrelmiller, wondering what your thoughts are regarding the `ConsistencyLevel: eventual` header. I feel that it would have been better to have something like: `AllowAdvancedQueries: true` in the HTTP header and Graph Service translating that into the replication model they use in the backend.

ddyett (approved):

ddyett:
> I really wish this database replication concept (eventual consistency) didn't go all the way to the SDK surface area, but this PR makes it one step closer to hiding that information from customers.
> 
> @darrelmiller, wondering what your thoughts are regarding the `ConsistencyLevel: eventual` header. I feel that it would have been better to have something like: `AllowAdvancedQueries: true` in the HTTP header and Graph Service translating that into the replication model they use in the backend.

i'm pretty sure there's been lively discussion about this one :)

MIchaelMainer (approved):
For ref

https://github.com/oasis-tcs/odata-vocabularies/blob/313294042be3f62ddd046ca8516a19ed447db11d/vocabularies/Org.OData.Capabilities.V1.md#readrestrictionsbase

PR #40, opened 2020-10-20
Update downloadDiff.ps1 to fix read-only variable

MIchaelMainer:
The `agent.jobstatus` environment variable has been set to read only (breaking change for us). We need to this pipeline to error when there are no changes to the metadata so that subsequent pipelines don't consider this pipeline a successful completion.

```log
##[error]Unable to process command '##vso[task.setvariable variable=agent.jobstatus;]canceled' successfully. Please reference documentation (http://go.microsoft.com/fwlink/?LinkId=817296)
##[error]Overwriting readonly variable 'agent.jobstatus' is not permitted. See https://github.com/microsoft/azure-pipelines-yaml/blob/master/design/readonly-variables.md for details.
```

- [ ] re-enable tasks for (v1.0 - 1) msgraph-capture-metadata after this PR.

### DIFF
--- a/scripts/downloadDiff.ps1
+++ b/scripts/downloadDiff.ps1
@@ -7,9 +7,12 @@
 #
 #.Parameter endpointVersion
 #   Specifies the metadata endpoint to target. Expected values are "v1.0" and "beta"
+#.Parameter targetBranch
+#   Optional. Specifies the branch to target. The default value is 'master'.
 #>
 
-param([parameter(Mandatory = $true)][String]$endpointVersion)
+param([parameter(Mandatory = $true)]  [String] $endpointVersion,
+      [parameter(Mandatory = $false)] [String] $targetBranch = 'master')
 
 # Contains Format-Xml function.
 Import-Module .\scripts\utility.ps1 -Force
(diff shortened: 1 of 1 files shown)


nikithauc:
Is this failure in the check the expected one?

zengin:
@MIchaelMainer I believe this is going to block docs-only changes like this one: https://github.com/microsoftgraph/msgraph-sdk-dotnet/pull/826, which is OK. I just want to double check that this is a deliberate decision.

MIchaelMainer:
@nikithauc I'm checking whether this is expected. I'm having a hard time recreating the errors locally.
@zengin You are correct, this won't pickup doc only changes. I'm okay with that. We never get issues based on the code comments. They will get picked at the next run.

zengin (approved):

nikithauc (approved):

MIchaelMainer:
@nikithauc One of the errors is the expected one. Another one was likely caused by me running the pipeline on a branch that was not targeting master (the one used in this PR). This script is tightly-bound to the master branch. Let me remedy that in this PR. Once this PR is merged, I can run on master and that should guide me on whether this change works as expected.

MIchaelMainer:
@nikithauc I've made an update that allows for branches other than master.

nikithauc:
> @nikithauc I've made an update that allows for branches other than master.

@MIchaelMainer  Thank you for getting the checks to run successfully!

PR #47, opened 2021-01-06
Metadata parser tool

zengin:
- Add a top level program that simply attempts to parse given metadata in args[0].
- Add default gitignore for dotnet


baywet (approved):
Since we now have a project with a dependency, can we add a dependabot configuration as well please?

baywet:
Also maybe a GitHub actions workflows to ci and validate pull requests. This way we have a short feedback loop and everything is at one place (I saw the change in the generator)

zengin:
@baywet, thanks for the suggestions. Added the simple build validation for the new tool as well as the dependabot config.

MIchaelMainer (approved):

.github/workflows/metadata-parser-build-validation.yml
```diff
@@ -0,0 +1,25 @@
...
+      with:
+        submodules: recursive
+    - name: Setup .NET
+      uses: actions/setup-dotnet@v1
+      with:
+        dotnet-version: 5.0.x
+    - name: Restore dependencies
+      run: dotnet restore ./tools/MetadataParser/MetadataParser.csproj
+    - name: Build
+      run: dotnet build ./tools/MetadataParser/MetadataParser.csproj --no-restore
```
baywet, line 25: We could run it for beta and for v1 as well to get feedback on the metadata changes

zengin: well, those files are auto generated and shouldn't be manually updated. And I have already added a step in generation pipeline to block checking in if it fails.

baywet: I was thinking about the case where somebody introduces a new transformation in a PR

zengin: I checked the last 10 XSLT changes, nobody has updated the clean metadata together with those changes, updates were picked up in subsequent full generations.

I also thought about putting this as a gate here and kicking off a PR from the full generation pipeline instead of directly checking into master, but that changes dependencies and structure of the pipeline a lot.

zengin: Maybe we can consider running it for the test metadata which is often updated with XSLT.

zengin: I have just run it against the test metadata, and it failed to parse :D

baywet: git blame! :D
I don't want to be annoying, it's just that I don't trust myself. The faster a machine can check my work and tell me I forgot something/did something wrong, the better :)

PR #48, opened 2021-01-07
Fix empty base type

MIchaelMainer:
And I expected to just see the 11 changes where `BaseType=""` is removed. There's much more than that - and I got a BSOD in the meantime so I lost the command I used to generate this last time. I see mostly doc updates, but also the removal of duplicate custom annotations like DerivedTypeConstraint. I also see some additions and there are certainly in the metadata. I may have used an older version of the metadata to generate against initially, or because of the dedup DerivedTypeConstraint, it may have been an already cleaned form of the metadata.

`.\typewriter.exe -v Info -m D:\repos\msgraph-metadata\beta_metadata.xml -o output -g TransformWithDocs -t D:\repos\msgraph-metadata\transforms\csdl\preprocess_csdl.xsl -d D:\repos\microsoft-graph-docs\ -e beta`


zengin (approved):
Looks good to me. Checked with the new tool that this parses fine as Edm model.

MIchaelMainer:
Pushed these changes I ran through typewriter with the doc fixes: https://github.com/microsoftgraph/microsoft-graph-docs/pull/11229/files

image

baywet (approved):

Issue #49, opened 2021-01-07
undocumented mandatory startDateTime/endDateTime query params on /calendarView


ivy-rew:
startDateTime & endDateTime not documented, but the service complains with 400 if their not present
https://docs.microsoft.com/en-us/graph/api/calendar-list-calendarview?view=graph-rest-1.0&tabs=http

I'm opening this issue here in hope that this is the feed for the api being shown via:
https://graphexplorerapi.azurewebsites.net/openapi?tags=me.calendar&openApiVersion=2&graphVersion=v1.0&format=json&style=PowerPlatform

A previous attempt to contribute is here: https://github.com/microsoftgraph/microsoft-graph-openapi/issues/12
 
AB#7459

PR #50, opened 2021-01-07
msgraph-meta #49: mandatory queryParams on /calendarView

ivy-rew:
… /calendarView


baywet:
@ivy-rew Thanks for the contribution! It's really appreciated.

@MIchaelMainer @zengin isn't this file supposed to be auto-generated? Most likely pulled from https://graphexplorerapi.azurewebsites.net/openapi?graphversion=v1.0&operationids=*&openapiversion=3 ?? It looks like it hasn't been updated in a while though.

MIchaelMainer:
@baywet Yes, I think it is generated with https://github.com/microsoftgraph/msgraph-metadata/tree/master/tools/odata2openapi. @peombwa, how do you handle scenarios like this for PowerShell? Would it be possible to add this as an OpenAPI extension?

ivy-rew:
I see that it doesn't make much sense to contribute change in a generated file. 
However, if there is a way to contribute these undocumented query params I'd gladly do so. 
Meanwhile I recon, that the **/user/{calendar-id}/calendar/calendarView** APIs are better documented. (only date date-format schema sub-type is not set there).

Even so, I use the /user resources instead of /me as workaround. Nevertheless, still it would be best to document these start-/endDateTime in a consistent way only once: and re-use them on all APIs that offer these capabilities.

baywet:
@ivy-rew This OpenAPI description is updated by the tool that Michael linked.
However I'm not sure why we have a fully built tool here and not a pointed to the sources instead. (having a built tool requires manual updates, makes the repo heavier...)
@MIchaelMainer do you happen to know where are the sources for that tool?

MIchaelMainer:
@baywet I'm sorry,  I don't where the source for that tool is maintained.

peombwa:
@MIchaelMainer @baywet We do use `CustomQueryOptions` annotation to add query params to paths. This is all done at the XSLT transform layer like this https://github.com/microsoftgraph/msgraph-metadata/pull/18.

StartDateTime/endDateTime query param is currently applied to `/calendarView` paths of users, groups, and bookingBusinesses. This can be accessed via https://graphexplorerapi.azurewebsites.net/openapi?graphversion=v1.0&operationids=*&openapiversion=3. We may need to add StartDateTime/endDateTime query params to `/me` as well. This wasn't implemented since we don't support `/me` in PowerShell.

baywet:
Thanks for jumping in @peombwa . So, is this open API description used at all then?

peombwa:
@baywet We do use it generate *calendarView cmdlets in PowerShell that have `StartDateTime/EndDateTime` as parameters https://github.com/microsoftgraph/msgraph-sdk-powershell/blob/dev/src/Calendar/Calendar/docs/v1.0/Get-MgUserCalendarView.md#-enddatetime.

baywet:
Hi @ivy-rew 
Thanks for your patience. I just had a call with @peombwa to understand the context better and list out the next steps.
This file is outdated and not used anymore, for your generation, you should use the API we pointed at in the last comments.

Now, that doesn't mean we don't want your PR, to the contrary! Can you make the following changes please: 
- [ ] delete the `openapi/v1.0/openapi.yaml`
- [ ] delete the `openapi/beta/openapi.yaml`
- [ ] delete the directory `tools/odata2openapi`
- [ ] update the XLST so it also applies the transformation for the `me` Singleton, add a [new matching rule ](https://github.com/microsoftgraph/msgraph-metadata/blob/master/transforms/csdl/preprocess_csdl.xsl#L404-L407)

[Here](https://github.com/microsoftgraph/msgraph-metadata/tree/master/transforms/csdl#command-line-instructions-for-running-the-transform) is some documentation on how to run the transform process. And don't hesitate to reach out if you have any question!

Thanks again a lot for your contributions!

baywet (changes requested):
list of changes provided in the previous comment

ivy-rew:
@baywet Thanks a lot for gathering all this information to supply a new fix.
I'm am amazed how you treat requests from the community. Actually, it inspires me to re-think my own attitude when reviewing code and guiding new members. 

Even so, I'm running quit of time to investigate in the feature behind this change request. In addition, though I do have some experience with XSLT, it's not an area where I feel comfortable :-) . So, I can't promise to actually file a new PR. But let it open till Saturday and let's see whether I can contribute something till then.

baywet:
thanks for the kind words! Give it a try and if it doesn't work (XSLT can be far from user friendly when not "used" to it), just apply the deletions, I'll push the XSLT change to your branch.

ivy-rew:
That sounds more than fair @baywet 
I've applied the suggested deletions. Feel free to take over :+1:

darrelmiller:
If we regenerate the OpenAPI, those parameters should be there.  Give me a few mins.

baywet:
Alright everyone, after chatting with @darrelmiller it appears the open API descriptions are still needed. Sorry about the confusion.
I authored #52 reusing some of your commits, the reasons why I had to put together another PR are:
- I don't have the permissions to push downstream
- I needed to test the new github actions workflow I added, and github token are not available for PRs from forks

This new PR adds a workflow that will keep the open API descriptions up to date, so you can keep using the files here instead of the graph explorer API. It also adds the missing XSLT pattern for the me calendar view.

Once the PR is merged, the CSDL will be updated (probably next week) and deployed to the GE API, which in turns should trigger an update of the open api descriptions here.
Closing this PR

PR #51, opened 2021-01-08
Add a parser validation for preprocessor test output

zengin:
- Addressed this comment: https://github.com/microsoftgraph/msgraph-metadata/pull/47#discussion_r553578286
- Removed a rule that is no longer applicable (there is no `settings` singleton in both V1 and Beta)


baywet (approved):
thanks for taking the time to add the suggestion

PR #53, opened 2021-01-15
Fix signatures of Accept, Decline and TentativelyAccept actions

### LINKED ISSUE #21: XSLT transform drops ProposedNewTime parameter
XSLT transform doesn't copy `ProposedNewTime` parameter while reordering of parameters alphabetically for `tentativelyAccept` and `decline` actions.

Assumption is that `ProposedNewTime` parameter didn't exist in metadata when the reordering rule was defined. 
[AB#7456](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/7456)

zengin:
Previous XSLT rule was replacing accept, decline and tentativelyAccept actions with a parameter reordering. The reason for the rule was to keep the code backwards compatible.

Our long term solution should be to have a parameters object which can extend freely with nullable properties.

The change I propose in this PR is changing the "replace" rule with an "add" rule, which adds the backwards compatible signature. Since I remove the replace rule, the original signatures from metadata gets copied as is.

See type summary comparisons between the following two files.
[typeSummary_accept_decline_base.txt](https://github.com/microsoftgraph/msgraph-metadata/files/5823336/typeSummary_accept_decline_base.txt)
[typeSummary_accept_decline_new.txt](https://github.com/microsoftgraph/msgraph-metadata/files/5823338/typeSummary_accept_decline_new.txt)

---
#### Through backwards compatible "replace" rules from XSLT
```
accept(Comment, SendResponse)
decline(Comment, SendResponse)
tentativelyAccept(Comment, SendResponse)
```
#### No compatible signatures with the actual metadata and docs
---
#### Through backwards compatible "add" rules from XSLT
```
accept(Comment, SendResponse)
decline(Comment, SendResponse)
tentativelyAccept(Comment, SendResponse)
```
#### By keeping signatures from original metadata
```
accept(SendResponse, Comment)
decline(ProposedNewTime, SendResponse, Comment)
tentativelyAccept(ProposedNewTime, SendResponse, Comment)
```
---

I have updated the test input XML file to match the actual metadata and updated output XML by running the new transform.

Fixes #21 

Fixes 3 C# Beta and 3 C# V1 Raptor tests. @baywet, I think this will fix some Java tests too.

### DIFF
--- a/transforms/csdl/preprocess_csdl.xsl
+++ b/transforms/csdl/preprocess_csdl.xsl
@@ -279,6 +279,67 @@
     <xsl:template match="edm:Schema[@Namespace='microsoft.graph.callRecords']/edm:Function[@Name='getDirectRoutingCalls']"/>
     <xsl:template match="edm:Schema[@Namespace='microsoft.graph']/edm:Function[@Name='delta'][edm:Parameter[@Name='token']][edm:Parameter[@Type='Collection(graph.site)']]"/>
 
+    <!-- Reorder action parameters -->
+
+    <!-- These actions have the same parameters that need reordering. Will need to create a new template
+         for each reordering. -->
+    <xsl:template match="edm:Schema[@Namespace='microsoft.graph']">
+      <xsl:copy>
+        <xsl:apply-templates select="@* | node()"/>
+        <xsl:element name="Action">
+          <xsl:attribute name="Name">accept</xsl:attribute>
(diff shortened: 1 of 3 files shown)


nikithauc (approved):

baywet (approved):
@zengin this is why this thing was always complaining! Thanks for looking into it.

PR #55, opened 2021-01-19
Update phoneMethods property to use PUT method

### LINKED ISSUE #54: Wrong method used for Update Phone Method for MFA
Hello,
According to https://docs.microsoft.com/en-us/graph/api/phoneauthenticationmethod-update?view=graph-rest-beta&tabs=http&viewFallbackFrom=graph-rest-1.0 the method used to update the phone number is PUT but the current code is using PATCH method:
https://github.com/microsoftgraph/msgraph-metadata/blob/master/openapi/beta/openapi.yaml#L391776 (for the /me) and
https://github.com/microsoftgraph/msgraph-metadata/blob/master/openapi/beta/openapi.yaml#L634534 for /users)

I don't know if I can just do a PR for that or if it's just a generated file I should have to modify by myself.
Thank you :)

jbpaux:
Fix #54
According to https://docs.microsoft.com/en-us/graph/api/phoneauthenticationmethod-update?view=graph-rest-beta&tabs=http the method is PUT not PATCH.
This PR intends to fix this behavior.


baywet:
Salut Jb!
Thanks for your contribution, we really appreciate it.
This file is being automatically generated, so we don't take manual changes for it.
However, I believe the service should be using PATH instead of PUT and is not following the OData conventions.
@darrelmiller any thoughts?

darrelmiller:
It was initially created as PUT and we provided guidance to the team that PATCH should be supported.  My understanding was that they were going to be doing that.  I will follow up with the PM.

jbpaux:
I close this one as I hope it will be fixed in other ways soon...

PR #57, opened 2021-01-26
Revert metadata to last known good state

zengin:
Non-fatal typewriter error in generation pipeline pushed changes to the metadata snapshots without validating with a smoke test in .NET SDK. Reverting to last known good state from Dec 8.


MIchaelMainer (approved):
related: https://github.com/microsoftgraph/MSGraph-SDK-Code-Generator/pull/417

PR #63, opened 2021-01-28
OpenAPI conversion validation as a PR gate

### LINKED ISSUE #62: XSLT rules can break Edm or OpenAPI model parsing
Our checkins for XSLT rules are not blocked by a PR gate that checks whether our transformations are still producing parsable outputs.

- Typewriter uses EDM model parsing
- Powershell uses OpenAPI conversion 
[AB#7832](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/7832)

zengin:
- adds OpenAPI reader as a submodule
- adds a step to MetadataParser tool to convert to OpenAPI
- adds a Github Action which will
  - apply the latest XSLT rules on the latest v1 and beta metadata
  - validate those transformed metadata is parsable

Fixes #62

### DIFF
--- a/scripts/run-metadata-validation.ps1
+++ b/scripts/run-metadata-validation.ps1
@@ -1,0 +1,48 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+<#
+.Synopsis
+    Runs MetadataParser tool to validate latest XSLT rules don't break downstream parsing of the metadata
+
+.Description
+    1. tranforms latests snapshots of beta and v1 metadata
+    2. validates that the transformed metadata is parsable as OData EDM model and conversion to OpenAPI document is possible
+
+.Example
+    ./scripts/run-metadata-validation.ps1 -repoDirectory C:/github/msgraph-metadata
+
+.Example
+    ./scripts/run-metadata-validation.ps1 -repoDirectory $GITHUB_WORKSPACE
+
+.Parameter repoDirectory
+    Full path the the root directory of msgraph-metadata checkout.
+#>
+
+param(
+    [Parameter(Mandatory=$true)][string]$repoDirectory
+)
+
+$transformCsdlDirectory = Join-Path $repoDirectory "transforms/csdl"
+$transformScript = Join-Path $transformCsdlDirectory "transform.ps1"
+$xsltPath = Join-Path $transformCsdlDirectory "preprocess_csdl.xsl"
+
+$betaSnapshot = Join-Path $repoDirectory "beta_metadata.xml"
+$v1Snapshot = Join-Path $repoDirectory "v1.0_metadata.xml"
+
+$metadataParserTool = Join-Path $repoDirectory "tools/MetadataParser/MetadataParser.csproj"
+
+$transformedBeta = Join-Path $repoDirectory "transformed_beta_metadata.xml"
+$transformedV1 = Join-Path $repoDirectory "transformed_v1.0_metadata.xml"
+
+Write-Host "Tranforming beta metadata using xslt..."
+& $transformScript -xslPath $xsltPath -inputPath $betaSnapshot -outputPath $transformedBeta
+
+Write-Host "Validating beta metadata after the transform..."
+& dotnet run -p $metadataParserTool $transformedBeta
+
+Write-Host "Tranforming v1.0 metadata using xslt..."
+& $transformScript -xslPath $xsltPath -inputPath $v1Snapshot -outputPath $transformedV1
+
+Write-Host "Validating v1.0 metadata after the transform..."
+& dotnet run -p $metadataParserTool $transformedBeta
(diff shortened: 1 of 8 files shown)


zengin:
Tested the new validation rule with the broken XSLT and it failed. Reverted back that change and it succeeded.

zengin:
@baywet can you have a look at the Github Action failure for updating latest OpenAPI descriptions? Not sure why the push fails.

tools/MetadataParser/Program.cs
```diff
@@ -1,12 +1,35 @@
 ﻿using Microsoft.OData.Edm.Csdl;
 using System.Xml;
 using System;
+using Microsoft.OData.Edm;
```
MIchaelMainer, line 4: nit: file should contain copyright header

MIchaelMainer (approved):

nikithauc (approved):

scripts/run-metadata-validation.ps1
```diff
@@ -0,0 +1,48 @@
...
+
+.Parameter repoDirectory
+    Full path the the root directory of msgraph-metadata checkout.
+#>
+
+param(
+    [Parameter(Mandatory=$true)][string]$repoDirectory
+)
+
+$transformCsdlDirectory = Join-Path $repoDirectory "transforms/csdl"
```
baywet, line 26:
```suggestion
$transformCsdlDirectory = Join-Path $repoDirectory "transforms" "csdl"
```

scripts/run-metadata-validation.ps1
```diff
@@ -0,0 +1,48 @@
...
+$transformScript = Join-Path $transformCsdlDirectory "transform.ps1"
+$xsltPath = Join-Path $transformCsdlDirectory "preprocess_csdl.xsl"
+
+$betaSnapshot = Join-Path $repoDirectory "beta_metadata.xml"
+$v1Snapshot = Join-Path $repoDirectory "v1.0_metadata.xml"
+
+$metadataParserTool = Join-Path $repoDirectory "tools/MetadataParser/MetadataParser.csproj"
```
baywet, line 33:
```suggestion
$metadataParserTool = Join-Path $repoDirectory "tools" "MetadataParser" "MetadataParser.csproj"
```

scripts/run-metadata-validation.ps1
```diff
@@ -0,0 +1,48 @@
...
+
+$transformedBeta = Join-Path $repoDirectory "transformed_beta_metadata.xml"
+$transformedV1 = Join-Path $repoDirectory "transformed_v1.0_metadata.xml"
+
+Write-Host "Tranforming beta metadata using xslt..."
+& $transformScript -xslPath $xsltPath -inputPath $betaSnapshot -outputPath $transformedBeta
+
+Write-Host "Validating beta metadata after the transform..."
+& dotnet run -p $metadataParserTool $transformedBeta
```
baywet, line 42: nit: I'd move the dotnet run steps in the workflow, this way we'd have a better integration for linting, warning etc... with the UI

baywet (changes requested):

baywet:
@zengin fixed the pipeline (I used your PR as guinea pig)!

peombwa (approved):

scripts/run-metadata-validation.ps1
zengin: Unfortunately `Join-Path` doesn't work with varargs and you need [weird tricks in piping](http://codebuckets.com/2016/08/20/powershell-join-path-with-multiple-parameters/) or multiple levels of function composition to get it working in single line.

Forward slash works cross platform.

scripts/run-metadata-validation.ps1
zengin: See other comment

scripts/run-metadata-validation.ps1
zengin: I would like to be able to run this script as a self contained test in local dev box without needing to create a PR and then realizing that my XSLT is wrong. That is the reason I am keeping everything in a single ps1 file. I wish Actions had a way to run as is on a local dev environment without much modifications.

zengin:
@baywet I added a quick fix to allow empty commits in OpenAPI action. Thanks for the other fixes you pushed.

I have addressed your comments. Can you please have another look?

scripts/run-metadata-validation.ps1
baywet: yeah not having the ability to run workflows locally is a real pain. I pushed 18 commits to fix the workflow before squashing and force push....

you could have two scripts:
- one that calculates the paths and settings them as tasks output (this one
- one that calls the first one and dotnet run for the local debug experience

zengin: I think this is a script that is short enough to be single file. And local testing instruction is as simple as:
> After you made your changes to the XSLT, run:
> `./scripts/run-metadata-validation.ps1 <local_metadata_checkout_folder>`

And everything is maintained in a single place.

baywet (approved):

PR #64, opened 2021-02-01
Remove reordering rule for event actions

zengin:
XLST rules were for backwards compatibility (i.e. to prevent breaking changes). Removing it will propagate what is in the original metadata.

This will cause a breaking change in C# and Java SDKs, so it is planned to be applied for Java V3 and C# V4 major releases.
Related #21

### DIFF
--- a/transforms/csdl/preprocess_csdl.xsl
+++ b/transforms/csdl/preprocess_csdl.xsl
@@ -464,30 +464,6 @@
     <xsl:template match="edm:Schema[@Namespace='microsoft.graph']/edm:Function[@Name='delta'][edm:Parameter[@Name='token']][edm:Parameter[@Type='Collection(graph.site)']]"/>
     <xsl:template match="edm:Schema[@Namespace='microsoft.graph']/edm:Function[@Name='additionalAccess'][edm:Parameter[@Name='accessPackageId']][edm:Parameter[@Type='Collection(graph.accessPackageAssignment)']][1]"/>
 
-    <!-- Reorder action parameters -->
-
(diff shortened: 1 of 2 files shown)


mairissi:
No impact for the Java SDK since as of version v3 we have parameter set.
It will be applied for C# v4.

darrelmiller:
@zengin  Removing the reorder rule would prevent us from guaranteeing a consistent order.  The way Exchange generate their CSDL can't guarantee the order.  We may end up breaking again in the future.  We need to move to parameter objects in C# before this rule can be removed.

zengin:
@darrelmiller this is not only a reordering rule as the actions have more parameters now and the "reordering" rule removes the additional parameters.

If the decision is to delay this until the parameter object is implemented, I would like to propose adding the missing parameters as part of C# major release, so that at least we support the functionality with the current list of parameters.

MIchaelMainer:
@mairissi  This isn't yet in C# Microsoft.Graph v4.

baywet:
reviewing from the kiota checklist: kiota based generation would solve for this for all languages:

- for actions we effectively generate a request body object, adding or reordering properties would have no impact
- for functions, we do use method parameters with building the request but those are ordered deterministically by the conversion process. Adding an override with a different set of parameters would generate a new request builder + namespace. Reordering the parameters of a function should be considered a breaking change at the API level (since the API call changes).

zengin:
Ready for review after 2 years :)

baywet (approved):

MIchaelMainer (approved):

PR #74, opened 2021-05-28
Add ProposedNewTime parameter to Decline and TentativelyAccept actions 

MIchaelMainer:
This adds the missing ProposedNewTime functionality for .NET and Java. We have to move the ProposedNewTime parameter to the end so not to break existing clients (.NET only).

We will remove the reorder rule when we generate parameter objects with Kiota.

Related to https://github.com/microsoftgraph/msgraph-metadata/pull/64

We should see a PR with the ProposedNewTime argument added to:
https://github.com/microsoftgraph/msgraph-sdk-dotnet/blob/dev/src/Microsoft.Graph/Generated/requests/EventDeclineRequestBuilder.cs

TODO
- [ ] why is this transform change not getting picked up in [generation](https://microsoftgraph.visualstudio.com/Graph%20Developer%20Experiences/_build/results?buildId=50835&view=results).

### DIFF
--- a/transforms/csdl/preprocess_csdl.xsl
+++ b/transforms/csdl/preprocess_csdl.xsl
@@ -300,14 +300,23 @@
 
     <!-- These actions have the same parameters that need reordering. Will need to create a new template
          for each reordering. -->
-    <xsl:template match="edm:Schema[@Namespace='microsoft.graph']/edm:Action[@Name='accept'][.//edm:Parameter[@Name='bindingParameter'][@Type='graph.event']]|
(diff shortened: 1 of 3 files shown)


mairissi (approved):

zengin:
I am not following why we need this change. I thought we agreed on releasing the original metadata in V4?

nikithauc (approved):

MIchaelMainer:
@zengin We are trying to maintain the same signature between v3 and v4 with the addition of ProposeNewTime. We will break this in vFuture with Kiota where we will remove the reordering. The parameter object will be added with Kiota.

PR #75, opened 2021-06-20
fix(metadata): user.memberOf, group.creationOptions,resourceBehaviorOptions,resourceProvisioningOptions

jgburet:
I'm coming from https://github.com/DefinitelyTyped/DefinitelyTyped/pull/53659

cf https://github.com/DefinitelyTyped/DefinitelyTyped/pull/53659/files#diff-d63698723579b98e65c314f7ad762c0031e80ad6e3b71cac6c252ee8ef2bae91R159 for what we receive from the API

### DIFF
--- a/v1.0_metadata.xml
+++ b/v1.0_metadata.xml
@@ -2503,7 +2503,7 @@
         <NavigationProperty Name="directReports" Type="Collection(graph.directoryObject)" />
         <NavigationProperty Name="licenseDetails" Type="Collection(graph.licenseDetails)" ContainsTarget="true" />
         <NavigationProperty Name="manager" Type="graph.directoryObject" />
-        <NavigationProperty Name="memberOf" Type="Collection(graph.directoryObject)" />
+        <NavigationProperty Name="memberOf" Type="Collection(graph.group)" />
         <NavigationProperty Name="oauth2PermissionGrants" Type="Collection(graph.oAuth2PermissionGrant)" />
         <NavigationProperty Name="ownedDevices" Type="Collection(graph.directoryObject)" />
         <NavigationProperty Name="ownedObjects" Type="Collection(graph.directoryObject)" />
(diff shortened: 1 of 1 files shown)


v1.0_metadata.xml
```diff
@@ -2503,7 +2503,7 @@
         <NavigationProperty Name="directReports" Type="Collection(graph.directoryObject)" />
         <NavigationProperty Name="licenseDetails" Type="Collection(graph.licenseDetails)" ContainsTarget="true" />
         <NavigationProperty Name="manager" Type="graph.directoryObject" />
-        <NavigationProperty Name="memberOf" Type="Collection(graph.directoryObject)" />
+        <NavigationProperty Name="memberOf" Type="Collection(graph.group)" />
```
baywet, line 2506: this should stay as it was. A user can be a member of a group, but also of a role. And this collection can return any of those, this is why it's using the parent type. If you want the collection to be filtered as groups (and down casted), you can append `/microsoft.graph.group` to your request.

v1.0_metadata.xml
```diff
@@ -3734,6 +3735,8 @@
         <Property Name="preferredLanguage" Type="Edm.String" />
         <Property Name="proxyAddresses" Type="Collection(Edm.String)" Nullable="false" />
         <Property Name="renewedDateTime" Type="Edm.DateTimeOffset" />
+        <Property Name="resourceBehaviorOptions" Type="Collection(Edm.String)" />
```
baywet, line 3738: this file is automatically generated from the metadata downloaded from the service, from the documentation information and from [an XSLT that applies transformations](https://github.com/microsoftgraph/msgraph-metadata/blob/master/transforms/csdl/preprocess_csdl.xsl). It'll be erased weekly.
I can see those properties are missing in the metadata even though they are returned by the service.

You should edit the XLST to insert those nodes automatically to to group entity, and you can test the transformation using the PowerShell script in the same directory.

The transformation will look something like
```xml 
<xsl:template match="edm:Schema[@Namespace='microsoft.graph']/edm:EntityType[@Name='group']">
   <xsl:copy>
       <xsl:copy-of select="@* | node()" />
       <Property Name="resourceBehaviorOptions" Type="Collection(Edm.String)" />
       <Property Name="resourceProvisioningOptions" Type="Collection(Edm.String)" />
       <Property Name="creationOptions" Type="Collection(Edm.String)" />
   </xsl:copy>
</xsl:template>
```

baywet (changes requested):
Thank you so much for your contribution!
I added a few comments to help finalize it. Don't hesitate if you have additional questions.

baywet:
closing due to inactivity

PR #76, opened 2021-07-08
Add navigation property restrictions transforms

irvinesunday:
Closes https://github.com/microsoftgraph/microsoft-graph-devx-api/issues/630

This PR:
- Adds `Org.OData.Capabilities.V1.NavigationRestrictions` annotations transforms for the below contained navigation properties
  - microsoft.graph.driveItem/workbook --> remove navigability
  - microsoft.graph.managedDevice/users --> remove indexability
  - microsoft.graph.user/joinedGroups --> remove indexability

We need to check their containment using these annotations as they are incorrect and we can't delete their containment value as this might lead to unintended consequences from other dependent tools. 

NB: The whitespace diff. is as a result of my editor automatically removing unnecessary whitespace on Save; I thought of reverting them but I figured it's an appropriate correction anyway.

PR #77, opened 2021-07-12
Add navigation property restrictions transforms

irvinesunday:
Closes https://github.com/microsoftgraph/microsoft-graph-devx-api/issues/630

This PR:
- Adds `Org.OData.Capabilities.V1.NavigationRestrictions` annotations transforms for the below contained navigation properties
  - microsoft.graph.driveItem/workbook --> remove navigability
  - microsoft.graph.managedDevice/users --> remove indexability
  - microsoft.graph.user/joinedGroups --> remove indexability

We need to check their containment using these annotations as they are incorrect and we can't delete their containment value as this might lead to unintended consequences from other dependent tools. 

Duplicate of this PR: https://github.com/microsoftgraph/msgraph-metadata/pull/76

### DIFF
--- a/transforms/csdl/preprocess_csdl.xsl
+++ b/transforms/csdl/preprocess_csdl.xsl
@@ -65,7 +65,7 @@
                   edm:Schema[@Namespace='microsoft.graph']/edm:EntityType[@Name='printer']/edm:NavigationProperty[@Name='allowedUsers']|
                   edm:Schema[@Namespace='microsoft.graph']/edm:EntityType[@Name='publishedResource']/edm:NavigationProperty[@Name='agentGroups']|
                   edm:Schema[@Namespace='microsoft.graph']/edm:EntityType[@Name='teamsApp']/edm:NavigationProperty[@Name='appDefinitions']|
-                  edm:Schema[@Namespace='microsoft.graph']/edm:EntityType[@Name='securityConfigurationTask']/edm:NavigationProperty[@Name='managedDevices']|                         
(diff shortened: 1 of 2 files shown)


baywet (approved):

transforms/csdl/preprocess_csdl.xsl
```diff
@@ -501,4 +501,62 @@
       </xsl:element>
     </xsl:copy>
   </xsl:template>
+  <xsl:template match="edm:Schema[@Namespace='microsoft.graph']">
```
andrueastman, line 504: Hey @baywet @irvinesunday.
It looks like this change is causing some tests to break on the SDK generator.

Should the match be applied indiscriminately? Or should it be more specific as seen [here](https://github.com/microsoftgraph/msgraph-metadata/blob/61269bb0c736df1276e66593173438efc2162d3a/transforms/csdl/preprocess_csdl.xsl#L469). 

The precondition assumed in the [test](https://github.com/microsoftgraph/MSGraph-SDK-Code-Generator/blob/dev/test/Typewriter.Test/Given_a_valid_metadata_file_to_Typewriter.cs#L560) will no longer apply as the cleanMetadata will have some capability annotations. This is irrespective of whether an input CSDL has the target present.

baywet: I'm not sure, to me adding a few global annotations should not impact the overall generation besides the targeted items.
That's unless some closing tag is missing/out of order...because you've got to love CSDL...

PR #99, opened 2021-11-08
New permissions metadata

darrelmiller:
Here is a PR of a prototype new format for modelling permissions for Microsoft Graph.  These files were based off the data used by Graph Explorer.


baywet:
Do you have a link to the spec it's following? We should probably define a JSON schema for it as well. (validation, versioning, etc...)

baywet:
closing this as it's ancient and conflicting

PR #101, opened 2021-11-10
Cleaning up Transform Rules 

ramsessanchez:
The transform rules in this PR are no longer useful as the ContainTarget has since been updated on these entities. We can dismiss these rules without affecting out metadata.


baywet:
I believe your local copy was behind when so started this PR.

```shell
git reset HEAD~1
git checkout *.yml
git switch master
git pull
git switch rsh/preprocessCleanUp
git rebase master
git push -f
```

PR #102, opened 2021-11-11
Cleanup Transform Rules

ramsessanchez:
The transform rules in this PR are no longer useful as the ContainTarget has since been updated on these entities. We can dismiss these rules without affecting out metadata.

Also including transformation rule for entity currently causing a build failure in the latest Beta Generation: https://github.com/microsoftgraph/msgraph-beta-sdk-java/pull/200


baywet (approved):

MIchaelMainer (approved):
Thank you for doing a cleanup pass on these.

baywet (approved):

MIchaelMainer (approved):

Issue #104, opened 2021-12-03
Add Custom XML Attribute to Limit Which Nav Properties Are Expanded


peombwa:
![image](https://user-images.githubusercontent.com/7061532/144636537-39c20fec-371d-4d5f-a7d4-60fe6a005575.png)

The attribute will be consumed by OpenAPI.NET.OData lib to limit the paths that are expanded. If a navigation property has `Explode="false"` and `ContainsTarget="true"`, OpenAPI.NET.OData will only generate paths to access the nav property (collection and by-key). All paths after the nav property will be dropped.

PR #105, opened 2021-12-03
Add `Explode` attribute to specific nav properties

peombwa:
This PR implements https://github.com/microsoftgraph/msgraph-metadata/issues/104 by adding an `Explode` attribute to targeted navigation properties that we don't want OpenAPI.Net.OData lib to expand.

Discussion on this attribute can be found in the `OpenAPI expanded paths` channel.

This change shouldn't affect typewriter since this is a custom attribute.

### DIFF
--- a/transforms/csdl/preprocess_csdl.xsl
+++ b/transforms/csdl/preprocess_csdl.xsl
@@ -634,4 +634,16 @@
       <xsl:apply-templates select="node()"/>
     </xsl:copy>
   </xsl:template>
+
+  <!-- Add custom 'explode' attribute -->
+  <xsl:template match="edm:Schema[@Namespace='microsoft.graph']/edm:EntityType[@Name='drive']/edm:NavigationProperty[@Name='bundles']|
+                       edm:Schema[@Namespace='microsoft.graph']/edm:EntityType[@Name='drive']/edm:NavigationProperty[@Name='following']|
+                       edm:Schema[@Namespace='microsoft.graph']/edm:EntityType[@Name='drive']/edm:NavigationProperty[@Name='special']">
+    <xsl:copy>
+      <!-- Select all attributes, add Explode="false" attribute, apply to the current node. -->
+      <xsl:apply-templates select="@*"/>
+      <xsl:attribute name="Explode">false</xsl:attribute>
+      <xsl:apply-templates select="node()"/>
+    </xsl:copy>
+  </xsl:template>
 </xsl:stylesheet>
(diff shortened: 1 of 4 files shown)


MIchaelMainer (approved):

baywet (approved):

irvinesunday:
@peombwa hold off merging this PR as this will necessitate an update to the OpenAPI.NET.OData library in the CSDL to Edm model parsing logic in this section... https://github.com/microsoft/OpenAPI.NET.OData/blob/fa2fee8ecaa5a3af6ea25f1f5329e3bd0dcd82ca/src/OoasGui/MainForm.cs#L165
image
...to use a different overloaded function that allows ignoring unexpected attributes and elements in the schema: https://github.com/OData/odata.net/blob/a93a25ba46fb2b1cb73c6c5ff8eb651d4356edba/src/Microsoft.OData.Edm/Csdl/CsdlReader.cs#L101-L114
image

irvinesunday:
> @peombwa hold off merging this PR as this will necessitate an update to the OpenAPI.NET.OData library in the CSDL to Edm model parsing logic in this section... https://github.com/microsoft/OpenAPI.NET.OData/blob/fa2fee8ecaa5a3af6ea25f1f5329e3bd0dcd82ca/src/OoasGui/MainForm.cs#L165 image ...to use a different overloaded function that allows ignoring unexpected attributes and elements in the schema: https://github.com/OData/odata.net/blob/a93a25ba46fb2b1cb73c6c5ff8eb651d4356edba/src/Microsoft.OData.Edm/Csdl/CsdlReader.cs#L101-L114 image

@peombwa turns out that this is no longer a blocker. However, we need to namespace the attribute for it to be accepted. See this for example: https://github.com/microsoft/OpenAPI.NET.OData/pull/152/files#diff-6781e21e38b7655a48a97a8da6956c8ec6453c6e5c1cf6007d700aef11c8f19f

peombwa:
@irvinesunday, just to recap on this. Will `Explode` annotation be supported in OpenAPI.NET.OData v1.0.10 or did we settle for something else? If supported, should the `Explode` annotation be in `Org.OData.Capabilities.V1.NavigationRestrictions` namespace for the lib to interpret correctly?

Given the above questions, will having explode annotation as shown below be adequate for us to suppress navigation properties that we don't want to expand i.e., `/drives/{id}/special/*` and `/drives/{id}/bundles/*`?
``` xml
<EntityType Name="drive" BaseType="graph.baseItem">
  <Property Name="driveType" Type="Edm.String" />
  <NavigationProperty Name="special" Type="Collection(graph.driveItem)" ContainsTarget="true">
    <Annotation Term="Org.OData.Capabilities.V1.NavigationRestrictions">
      <Record>
        <PropertyValue Property="Explode" Bool="false" />
      </Record> 
    </Annotation>
  </NavigationProperty>
</EntityType>
```

PR #107, opened 2022-01-11
- fixes a bug where open api document update could be based on outdated cached version in devx API

baywet:


baywet:
Sorry, setting auto-merge from the CLI merged right away as there are no policies on this repo 🙏

PR #110, opened 2022-01-31
adds inner error description

baywet:
inserts a description for the inner error type as prepration work for Kiota.

We're not describing the error itself as it's brought in the conversion process by OData conventions.

related https://github.com/microsoft/kiota/issues/411
related https://github.com/microsoft/OpenAPI.NET.OData/pull/167

More reading https://docs.microsoft.com/en-us/graph/errors

(I believe the inner error description is partly wrong, as it doesn't inherit from error according to the odata spec)

### DIFF
--- a/transforms/csdl/preprocess_csdl.xsl
+++ b/transforms/csdl/preprocess_csdl.xsl
@@ -501,6 +501,38 @@
   <xsl:template match="edm:Schema[@Namespace='microsoft.graph']">
     <xsl:copy>
       <xsl:apply-templates select="@* | node()"/>
+      <xsl:choose>
+        <!-- Add inner error description -->
+        <xsl:when test="$add-innererror-description='True'">
+          <xsl:element name="ComplexType">
+            <xsl:attribute name="Name">InnerError</xsl:attribute>
+            <xsl:element name="Property">
+              <xsl:attribute name="Name">request-id</xsl:attribute>
+              <xsl:attribute name="Type">Edm.String</xsl:attribute>
+              <xsl:element name="Annotation">
+                <xsl:attribute name="Term">Org.OData.Core.V1.Description</xsl:attribute>
+                <xsl:attribute name="String">Request Id as tracked internally by the service</xsl:attribute>
+              </xsl:element>
+            </xsl:element>
+            <xsl:element name="Property">
+              <xsl:attribute name="Name">client-request-id</xsl:attribute>
+              <xsl:attribute name="Type">Edm.String</xsl:attribute>
+              <xsl:element name="Annotation">
+                <xsl:attribute name="Term">Org.OData.Core.V1.Description</xsl:attribute>
+                <xsl:attribute name="String">Client request Id as sent by the client application.</xsl:attribute>
+              </xsl:element>
+            </xsl:element>
+            <xsl:element name="Property">
+              <xsl:attribute name="Name">Date</xsl:attribute>
+              <xsl:attribute name="Type">Edm.DateTimeOffset</xsl:attribute>
+              <xsl:element name="Annotation">
+                <xsl:attribute name="Term">Org.OData.Core.V1.Description</xsl:attribute>
+                <xsl:attribute name="String">Date when the error occured.</xsl:attribute>
+              </xsl:element>
+            </xsl:element>
+          </xsl:element>
+        </xsl:when>
+      </xsl:choose>
       <!-- Remove navigability for workbook navigation property -->
       <xsl:element name="Annotations">
         <xsl:attribute name="Target">microsoft.graph.driveItem/workbook</xsl:attribute>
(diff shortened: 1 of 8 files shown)


zengin (changes requested):

transforms/csdl/preprocess_csdl.xsl
```diff
@@ -634,4 +634,38 @@
...
   </xsl:template>
+
+  <!-- Add inner error description -->
+  <xsl:template match="edm:Schema[@Namespace='microsoft.graph']">
+    <xsl:copy>
+      <xsl:apply-templates select="@* | node()"/>
+      <xsl:element name="ComplexType">
+        <xsl:attribute name="Name">InnerError</xsl:attribute>
+        <xsl:element name="Property">
+          <xsl:attribute name="Name">request-id</xsl:attribute>
```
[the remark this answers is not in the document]
baywet: thanks, I was afraid of that, I'm not sure who designed the inner error payload back then ignoring every convention...
Ideally we don't want this type in the TypeWriter based SDK, since it's already assumed in the generation process. I was thinking about adding a super custom attribute on the type (like TypeWriter:Ignore="true") and making a dirty patch over there to skip it. Thoughts?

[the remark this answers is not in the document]
zengin: Can we add a parameter similar to:
https://github.com/microsoftgraph/msgraph-metadata/blob/fd8d068cd284a5693be9d1052d5f91d0b54211d1/transforms/csdl/preprocess_csdl.xsl#L7

and then maybe produce two different clean metadata versions in the generation process, one for OpenAPI based SDKs and one for Typewriter SDKs.

I am OK with a temporary hack idea as well since Typewriter is going away, but the param approach could arguably take the same amount of time to implement.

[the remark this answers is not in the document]
baywet: good suggestion!
I just pushed a commit to add the parameter, let me know what you think.
Now the question I have is how to we "fork the description" in the generation pipeline? different build artifacts and run the scripts twice with different argument sets?

transforms/csdl/preprocess_csdl.xsl
```diff
@@ -5,6 +5,7 @@
     <xsl:output method="xml" indent="yes"/>
     <xsl:strip-space elements="*"/> <!-- Remove empty space after deletions. -->
     <xsl:param name="remove-capability-annotations">True</xsl:param>
+    <xsl:param name="remove-innererror-description">True</xsl:param>
```
MIchaelMainer, line 8: I suggest that we name this `add*` since that is what we are doing here. The default should be `False` since we want existing use of this transform to continue without change.

```suggestion
    <xsl:param name="add-innererror-description">False</xsl:param>
```

transforms/csdl/preprocess_csdl.xsl
```diff
@@ -634,4 +635,42 @@
       <xsl:apply-templates select="node()"/>
     </xsl:copy>
   </xsl:template>
+
+  <!-- Add inner error description -->
+  <xsl:template match="edm:Schema[@Namespace='microsoft.graph']">
+    <xsl:choose>
+      <xsl:when test="$remove-innererror-description='False'">
```
MIchaelMainer, line 642: Matches the comment and intention

```suggestion
      <xsl:when test="$add-innererror-description='True'">
```

transforms/csdl/transform.ps1
```diff
@@ -11,7 +11,9 @@ param (
     [bool]
     $dbg = $false,
     [bool]
-    $removeCapabilityAnnotations = $true
+    $removeCapabilityAnnotations = $true,
+    [bool]
+    $removeInnererrorDescription = $true
```
MIchaelMainer, line 16:
```suggestion
    $addInnerErrorDescription = $false
```

transforms/csdl/readme.md
```diff
@@ -24,10 +24,12 @@ Please use fully qualified and precise template match statements so as not to pe
     > NB: The build job in the pipeline currently errors out on PRs coming in from personal forked repos.  
 
 ## Command line instructions for running the transform
+
 1. Start PowerShell
 2. `cd` into `transforms\csdl` folder
-3. Run `transform.ps1 <xsl_file> <input_file> <output_file>`. If files are not specified, the script will apply transformations from  *preprocess_csdl.xsl* on  *preprocess_csdl_test_input.xml* and override *preprocess_csdl_test_output.xml* file. 
+3. Run `transform.ps1 <xsl_file> <input_file> <output_file>`. If files are not specified, the script will apply transformations from  *preprocess_csdl.xsl* on  *preprocess_csdl_test_input.xml* and override *preprocess_csdl_test_output.xml* file.
 4. You can optionally run the transform without removing capability annotations by setting the `-removeCapabilityAnnotations` flag to `$false`. The default behavior is to remove capability annotations.
+5. You can additional run the transform without injecting the inner error description by setting the `-removeInnererrorDescription` false to `$true` (default, false to include it).
```
MIchaelMainer, line 32:
```suggestion
5. You can optionally run the transform to inject the OData inner error description by setting the `addInnerErrorDescription` flag to `$true`. The default value is `$false`.
```

transforms/csdl/transform.ps1
```diff
@@ -36,17 +38,21 @@ $outputFullPath = Get-PathWithPrefix -requestedPath $outputPath
 
 $xsltargs = [System.Xml.Xsl.XsltArgumentList]::new()
 $xsltargs.AddParam("remove-capability-annotations", "", $removeCapabilityAnnotations.ToString())
+$xsltargs.AddParam("remove-innererror-description", "", $removeInnererrorDescription.ToString())
```
MIchaelMainer, line 41:
```suggestion
$xsltargs.AddParam("add-innererror-description", "", $removeInnererrorDescription.ToString())
```

MIchaelMainer (changes requested):
This new functionality should be optional so that existing use of the transform do not need to be updated.

transforms/csdl/preprocess_csdl.xsl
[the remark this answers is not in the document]
zengin: That's what I had in mind.

The only addition I have is that maybe we can call the parameter something like `transform-for-openapi` instead of making it specific to generation of error descriptions. That will give us the reusability of the same parameter for all other scenarios where the fix is OpenAPI specific. We can postpone this until we have additional scenarios come in too.

zengin (approved):

transforms/csdl/preprocess_csdl.xsl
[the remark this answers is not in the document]
baywet: I'd suggest we postpone it until we have additional use cases (like contains target), at least now the name of the parameter is explicit to what it does. resolving.

baywet:
@zengin do you have any idea why the workflow is failing at this point?

transforms/csdl/transform.ps1
```diff
@@ -36,17 +38,21 @@ $outputFullPath = Get-PathWithPrefix -requestedPath $outputPath
 
 $xsltargs = [System.Xml.Xsl.XsltArgumentList]::new()
 $xsltargs.AddParam("remove-capability-annotations", "", $removeCapabilityAnnotations.ToString())
+$xsltargs.AddParam("add-innererror-description", "", $removeInnererrorDescription.ToString())
```
zengin, line 41:
```suggestion
$xsltargs.AddParam("add-innererror-description", "", $addInnerErrorDescription.ToString())
```

zengin:
@baywet now the error looks like on how the transform is applied. The rule removes the whole microsoft.graph namespace for some reason.

baywet:
Maybe the chose tag should wrap around the template one? (not in front of my computer right now)
Or maybe we need and empty else statement?
Gee this brings my back to my SharePoint on prem dev days with search customization. I don't miss it. 😂

zengin (approved):

Issue #114, opened 2022-02-18
Use XSLT to add support for properly describing delta functions


darrelmiller:
Delta functions need to have annotations added to signal that they will return `@Odata.DeltaLink` and `@OData.NextLink` in the response payload.

We need to identify how to annotation the functions properly.

baywet:
Closing at this is added by convention by the conversion library. BaseDeltaFunctionResponse has the deltalink and nextlink properties

PR #115, opened 2022-02-20
Suggestion for correct description of the users delta response

avgdorom:
I've tried to use the Go SDK to fetch users delta data. Apparently, there's a bug in the specifications of the response to that call in the OpenAPI doc.
This is my (possibly lame) attempt to solve https://github.com/microsoftgraph/msgraph-metadata/issues/114.


baywet (changes requested):
@avgdorom Thanks for the contribution, and sorry we can't accept it in its current form either.
This is mainly because this file is auto-generated as well, and will be overwritten every week.
The API publishing process on Microsoft Graph is a bit complex, I'll try to share some lights here.
When a **workload** team comes to Microsoft Graph to publish a new API (let's say OneDrive team wants to publish a new photo tagging API), they suggest modifications to a main document that describes the whole API which leaves [here](https://graph.microsoft.com/v1.0/$metadata)
The SDK generation pipeline copies this file to a local version of this repo [here](https://github.com/microsoftgraph/msgraph-metadata/tree/master/clean_v10_metadata) and applies an [XLS transformation](https://github.com/microsoftgraph/msgraph-metadata/blob/master/transforms/csdl/preprocess_csdl.xsl) to perform minor fixes in this document so the SDK experience is better.

All of that relies on OData and CSDL and a lot of conventions.
Once the CSDL is cleaned up and enhanced with documentation, the weekly pipeline uses the [odata to open api conversion library](https://github.com/microsoft/OpenAPI.NET.OData) to generate and overwrite the document you've edited.

Now, if we zoom on this particular issue, Microsoft Graph is not using the $delta "keyword" defined by OData conventions, but a delta function, which relies on additional conventions defined by Microsoft Graph (on top of the OData ones)
The return type for all those delta functions is defined as a Collection(Type) (type being user, application, message...)
This should give is a object with a value property which is an array of those objects. But because of a bug in the conversion lib which has been fixed but not released yet, the conversion lib is not generating the right information in the OpenAPI description document.
[More context here](https://github.com/microsoftgraph/msgraph-sdk-go/issues/57). This will be fixed once the library is released.

For all the properties with an @odata prefix, they are call instance annotations and they are added at runtime by convention by the service. However they are not part of the CSDL description. odata context and odata nextlink will be added by convention by the conversion library with the fix I mentioned earlier.
We're left with odata deltaLink, which is not standard odata conventions.
I'm currently working with @darrelmiller and @mikepizzo, both API architects, to understand whether or not there's a way to describe that annotation in the CSDL so the conversion library can pick it up and include it in the OpenAPI description. Here is what we'll do after we get that information

If yes:
1. Update the XSLT to inject the missing information in the CSDL.
2. Rely on the conversion library to carry it forward in the OpenAPI description.
3. Fix the conversion library if it doesn't do the right thing.

If no:
1. Define some kind of post-processing capability for OpenAPI (XSLT for OData, probably something JMESPath based)
2. Build some kind of quick transformation engine.
3. Add rules for all the delta path items.
4. Run that transformation after the conversion.

I hope this lengthy explanation shine a light on why "this is so complicated" and I'm sorry again we're not in a position we can accept community contributions on that issue. Don't hesitate to ask follow up questions.

avgdorom:
@baywet Thank you for the explanation. Can't say I got it all, but the message that this is too complicated for me was conveyed perfectly 😄 
So, in that case I will wait for a fix from you. Got any ETA? Just a ballpark will do... 😃

baywet:
We're hoping to release the new version of the conversion lib in the next 2 weeks which should fix the shape of the return type. Once that is done, if we've been able to describe and generate a property for the delta link, it'll be there as well, if we haven't the property will be accessible in the additional data manager on the object, which should at least unblock you.

avgdorom:
@baywet Thank you!

baywet:
Sorry for dropping the ball on this one.
Two issues are logged in the conversion library
https://github.com/microsoft/OpenAPI.NET.OData/issues/196
https://github.com/microsoft/OpenAPI.NET.OData/issues/231

Closing this as you can follow the progress there.

PR #117, opened 2022-02-28
- switches to hidi for CSDL to openAPI conversion

baywet:
depends on hidi 0.6.0 preview 1 to be released ( @MaggieKimani1 will work on that tomorrow)

### DIFF
--- a/scripts/GenerateOpenApi.ps1
+++ b/scripts/GenerateOpenApi.ps1
@@ -11,8 +11,36 @@
 
 param([parameter(Mandatory = $true)][String]$endpointVersion)
 
-$url = "https://graphexplorerapi.azurewebsites.net/openapi?operationIds=*&openApiVersion=3&graphVersion=$endpointVersion&format=yaml&forceRefresh=true"
+$outputFile = Join-Path $PSScriptRoot ".." "openapi" $endpointVersion "openapi.yaml"
+$oldOutputFile = "$outputFile.old"
+$cleanVersion = $endpointVersion.Replace(".", "")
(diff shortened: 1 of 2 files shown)


scripts/GenerateOpenApi.ps1
```diff
@@ -11,8 +11,36 @@
...
+}
+
+try {
+    Invoke-Expression "hidi transform --csdl ""$inputFile"" --output ""$outputFile"" --version OpenApi3_0 --loglevel Information --format yaml"
+} catch {
+    if(Test-Path $oldOutputFile)
+    {
+        Write-Warning "Restoring old output file"
+        $originalFileName = Split-Path $outputFile -leaf
+        Rename-Item $oldOutputFile $originalFileName
```
MIchaelMainer, line 42: On L32, $outputFile, ..\openapi\version\openapi.yaml is renamed to ..\openapi\version\openapi.yaml.old, which is then used as input on L36.
L41, $originalFileName has a value of openapi.yaml.old
L42, $oldOutputFile has a value of ..\openapi\version\openapi.yaml.old, which is the same name as $originalFileName.

Am I reading this correctly? I also question if it matters if we restored the old input since if there was an error, we would stop processing on error I assume.

baywet: no, L41 $originalFileName has a value of openapi.yaml not openapi.yaml.old.

The reason why I added this rename/restore/delete mechanism is because hidi stops if the target file already exists instead of overwritting it. But if it fails during the transformation process I don't want to lose the original file. A cheap "backup" mechanism if you want since I can't guarantee the directory will be under source control and/or will support transactions.

MIchaelMainer (approved):

peombwa (approved):

PR #141, opened 2022-04-25
Remove insertability and deletability from site resources.

peombwa:
This PR removes insertability and deletability from site resources as described in https://docs.microsoft.com/en-us/graph/api/resources/sharepoint?view=graph-rest-beta:
> The SharePoint API in Microsoft Graph supports the following core scenarios:
> - Access to Read-only support for site resources (no ability to create new sites)
> ...

This is done to avoid confusion as we currently generate create and delete commands for a site that are not supported by the API.

### DIFF
--- a/transforms/csdl/preprocess_csdl.xsl
+++ b/transforms/csdl/preprocess_csdl.xsl
@@ -544,6 +544,7 @@
         </xsl:copy>
     </xsl:template>
 
+    <!-- Capability Annotations Templates -->
     <xsl:template name="DeleteRestrictionsTemplate">
         <xsl:param name = "deletable" />
         <xsl:element name="Annotation">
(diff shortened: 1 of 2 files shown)


MIchaelMainer (approved):

zengin (approved):

andrueastman:
Hey @peombwa 

It looks like this change is causing a breaking change on TypeWriter based SDKs. If its okay, can we revert this PR and reapply it to have the restrictions only apply to these files that are used for openAPI based generation?

- https://github.com/microsoftgraph/msgraph-metadata/blob/master/clean_v10_metadata/cleanMetadataWithDescriptionsAndAnnotationsAndErrorsv1.0.xml
- https://github.com/microsoftgraph/msgraph-metadata/blob/master/clean_v10_metadata/cleanMetadataWithDescriptionsAndAnnotationsAndErrorsv1.0.xml

I have created #142 to track.

peombwa:
@andrueastman, sorry about that. Please revert the change (already reverted). We will address this in https://github.com/microsoftgraph/msgraph-metadata/issues/142.

PR #143, opened 2022-04-27
Enable GetPstnCalls and GetDirectRoutingCalls in open Api generation

andrueastman:
This PR closes https://github.com/microsoftgraph/MSGraph-SDK-Code-Generator/issues/277

It enables the generation of the `GetPstnCalls` and `GetDirectRoutingCalls` routes in the open Api document to enable the routes in the SDK generation

### DIFF
--- a/transforms/csdl/preprocess_csdl.xsl
+++ b/transforms/csdl/preprocess_csdl.xsl
@@ -7,6 +7,14 @@
     <xsl:param name="remove-capability-annotations">True</xsl:param>
     <xsl:param name="add-innererror-description">False</xsl:param>
 
+    <!-- Flag to signal if we are generating a document for open api generation. -->
+    <xsl:variable name="open-api-generation">
+        <xsl:choose>
+            <!-- Open API document generation is done with capability annotations and error descriptions -->
+            <xsl:when test="$remove-capability-annotations='False' and $add-innererror-description='True'">True</xsl:when>
+            <xsl:otherwise>False</xsl:otherwise>
+        </xsl:choose>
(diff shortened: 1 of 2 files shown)


transforms/csdl/preprocess_csdl.xsl
```diff
@@ -334,8 +334,10 @@
     </xsl:template>
 
     <!--Remove functions that are blocking beta generation-->
-    <xsl:template match="edm:Schema[@Namespace='microsoft.graph.callRecords']/edm:Function[@Name='getPstnCalls']"/>
-    <xsl:template match="edm:Schema[@Namespace='microsoft.graph.callRecords']/edm:Function[@Name='getDirectRoutingCalls']"/>
+    <xsl:if test="$remove-capability-annotations='True'"><!--Remove functions that are blocking beta generation only for CSDL generation-->
```
zengin, line 337: I think we are overloading the meaning of "remove capability annotations". If it is not too costly, please introduce a new switch.

zengin (approved):

MIchaelMainer (commented):
Please add a test case and check if we can use another signal instead of **remove-capability-annotations**. 

Perhaps at this time we need a condition to capture changes we need in support of the OpenAPI specification to fence off from the changes in support typewriter.

transforms/csdl/preprocess_csdl.xsl
andrueastman: I hear you. I will update the PR again to reflect this.

baywet (approved):

PR #144, opened 2022-04-27
Remove insertability and deletability from site resource to only apply to openApi based generation

### LINKED ISSUE #142: Remove insertability and deletability from site resource to only apply to openApi based generation
https://github.com/microsoftgraph/msgraph-metadata/pull/141 looks to add delete restrictions however these have been applied to all the metadata files causing a breaking change in the metadata.

Ideally, we should add the restriction to only to these files so as not to affect TypeWriter based generation if we don't intend to have a breaking change in the SDKs.

- https://github.com/microsoftgraph/msgraph-metadata/blob/master/clean_v10_metadata/cleanMetadataWithDescriptionsAndAnnotationsAndErrorsv1.0.xml
- https://github.com/microsoftgraph/msgraph-metadata/blob/master/clean_v10_metadata/cleanMetadataWithDescriptionsAndAnnotationsAndErrorsv1.0.xml

andrueastman:
Closes #142

### DIFF
--- a/transforms/csdl/preprocess_csdl.xsl
+++ b/transforms/csdl/preprocess_csdl.xsl
@@ -692,19 +692,21 @@
                     <xsl:with-param name="deletable">false</xsl:with-param>
                 </xsl:call-template>
             </xsl:element>
-            <xsl:element name="Annotations">
-                <xsl:attribute name="Target">microsoft.graph.site</xsl:attribute>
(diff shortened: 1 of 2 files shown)


zengin (approved):

peombwa (approved):

andrueastman:
Thanks @zengin and @peombwa

Issue #150, opened 2022-05-10
Duplicate annotation beta


andrueastman:
This is related https://github.com/microsoftgraph/msgraph-metadata/issues/145

XSLT transform adds an already existing annotation causing beta generation fail

https://github.com/microsoftgraph/msgraph-metadata/blob/96a80bba7eb4f1792a41beae50b1509b3c2d2ece/transforms/csdl/preprocess_csdl.xsl#L624

This is related to recent changes made in apidoctor via https://github.com/OneDrive/apidoctor/pull/116

Ideally the transform should probably modify the xml section rather than add a duplicate

Issue #145, opened 2022-04-28
Annotation Duplication 


georgend:
https://github.com/microsoftgraph/msgraph-metadata/blob/3a5f4126532b32ce4e34e6afa4206156454ad04b/transforms/csdl/preprocess_csdl.xsl#L695-L708

The transform above cause the annotation to be spread out over two annotation tags. Such as:-
<img width="475" alt="image" src="https://user-images.githubusercontent.com/1641829/165730308-dc186057-4546-4d1a-badf-f39be9f29e56.png">

This breaks expected format whereby only one Annotations tag per target is expected. 
Can we collapse it into one Annotations tag such as:-
<img width="475" alt="image" src="https://user-images.githubusercontent.com/1641829/165730876-fa7677f4-f8b4-4801-a896-e7305124a1bb.png">

PR #151, opened 2022-05-10
Add the `microsoft.graph.user/joinedGroups` annotations otherwise modify it if exists

andrueastman:
This PR closes https://github.com/microsoftgraph/msgraph-metadata/issues/150

Related to #145 

It makes changes to the CSDL transform to add the `microsoft.graph.user/joinedGroups` annotations to cater for the v1.0 metadata. 

However, if the annotation already exists, we simply modify it by appending the subset of annotations.

This is done to prevent issues observed after work done through https://github.com/OneDrive/apidoctor/pull/116 where annotations targets should ideally be unique, and an XML document shouldn't have multiple annotation tags with the same target in the same doc.

### DIFF
--- a/transforms/csdl/preprocess_csdl.xsl
+++ b/transforms/csdl/preprocess_csdl.xsl
@@ -570,6 +570,25 @@
             </xsl:element>
         </xsl:element>
     </xsl:template>
+    <xsl:template name="NavigationRestrictionsTemplate">
+        <xsl:param name = "navigable" />
+        <xsl:element name="Annotation">
+            <xsl:attribute name="Term">Org.OData.Capabilities.V1.NavigationRestrictions</xsl:attribute>
+            <xsl:element name="Record" namespace="{namespace-uri()}">
+                <xsl:element name="PropertyValue">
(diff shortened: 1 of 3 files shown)


transforms/csdl/preprocess_csdl.xsl
```diff
@@ -708,6 +720,18 @@
         </xsl:copy>
     </xsl:template>
 
+    <!-- Remove indexability for joinedGroups navigation property -->
+    <!-- If the annotation already exists modify it -->
```
peombwa, line 724: Can we apply this logic to all existing nav properties to avoid future breaking changes?

peombwa (approved):

MIchaelMainer (approved):

transforms/csdl/preprocess_csdl.xsl
andrueastman: Thanks. Will resolve in next update.

andrueastman: Resolved in latest update.

peombwa (approved):

andrueastman:
Thanks @peombwa

PR #160, opened 2022-06-02
Fix template matching clause for `places` EntitySet

andrueastman:
This PR is part of https://github.com/microsoftgraph/msgraph-sdk-dotnet-core/issues/141

It addressed a bug in the XSLT transform that would incorrectly match the `places` EntitySet so as to add the relevant `DerivedTypeConstraint` annotation to the metadata as it failed to cater for the fact that the `EntitySet` is held within an `EntityContainer`.

The test input has been updated to validate that the fix works.


baywet (approved):

andrueastman:
Thanks @baywet

PR #171, opened 2022-06-23
Schema sync 2022-06-23 from 1.1.21

marabooy:
Schema sync 2022-06-23 from 1.1.21


zengin:
@marabooy, could you please rebase this PR? I think merging #170 caused the conflict.

And assuming these PRs are taking a snapshot from files elsewhere, can we assume the latest open PR will have the changes from the previously opened PRs? If that is the case, maybe we can close older outstanding PRs in favor of the latest one as a process on our end. What do you think?

marabooy:
Let me publish the new version in a new branch.

PR #173, opened 2022-06-28
Adds OpenAPI parsing as a gate

zengin:
- partial https://github.com/microsoftgraph/MSGraph-SDK-Code-Generator/issues/782

### DIFF
--- a/tools/OpenAPIParser/OpenAPIParser.csproj
+++ b/tools/OpenAPIParser/OpenAPIParser.csproj
@@ -1,0 +1,14 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net6.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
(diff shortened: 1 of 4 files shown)


zengin:
Test workflow run with broken OpenAPI doc: https://github.com/microsoftgraph/msgraph-metadata/actions/runs/2573308096

andrueastman (approved):

tools/OpenAPIParser/OpenAPIParser.csproj
```diff
@@ -0,0 +1,14 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
```
baywet, line 3: Why not use hidi with the validate command instead?

zengin: This (what kiota and DevX API do) was the quickest I could find at 7pm while dealing with the ICM and the time pressure to block today's generation with a gate :) I opened an issue to address this referencing this comment.

baywet: thanks for dealing with this! I do understand the urgency aspect of this PR.

PR #181, opened 2022-07-05
- fixes a bug where beta validation would always succeed

baywet:
because the script is running a second process after beta validation, the exit code would be overwritten, which meant beta would always go through

### DIFF
--- a/scripts/run-openapi-validation.ps1
+++ b/scripts/run-openapi-validation.ps1
@@ -30,5 +30,19 @@ $openAPIParserTool = Join-Path $repoDirectory "tools/OpenAPIParser/OpenAPIParser
 Write-Host "Validating beta OpenAPI doc..." -ForegroundColor Green
 & dotnet run --project $openAPIParserTool $betaYaml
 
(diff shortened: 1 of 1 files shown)


scripts/run-openapi-validation.ps1
```diff
@@ -30,5 +30,19 @@ $openAPIParserTool = Join-Path $repoDirectory "tools/OpenAPIParser/OpenAPIParser
...
 & dotnet run --project $openAPIParserTool $betaYaml
 
+$finalExitCode = 0
+
+if ($LASTEXITCODE -ne 0) {
+    Write-Error "Validation failed for beta OpenAPI doc"
+    $finalExitCode = 1
+}
+
 Write-Host "Validating beta OpenAPI doc..." -ForegroundColor Green
```
zengin, line 40: fixing my own typo :)
```suggestion
Write-Host "Validating v1 OpenAPI doc..." -ForegroundColor Green
```

zengin (approved):
@baywet thanks for catching and fixing this.

PR #190, opened 2022-07-27
Specification for Graph Permissions Model 

darrelmiller:


specs/permissions.md
```diff
@@ -0,0 +1,340 @@
...
+
+```
+(User.Read | User.Read.All) & Group.Read
+```
+
+### includedProperties
+The "includedProperties" member is an array of strings that identify properties of the resource representation returned by the path, that are accessible with the permission.
+
+### excludedProperties
+The "includedProperties" member is an array of strings that identify properties of the resource representation returned by the path, that are not accessible with the permission.
```
PetRich-MSFT, line 116:
```suggestion
The "excludedProperties" member is an array of strings that identify properties of the resource representation returned by the path, that are not accessible with the permission.
```

specs/permissions.md
```diff
@@ -0,0 +1,340 @@
...
+### requiresAdminConsent
+The "requiresAdminConsent" member is a boolean value with a default value of false. When true, this permission can only be consented by an adminstrator.
+
+## <a name="pathObject"></a>Path Object
+The path object contains properties that affect how the permission object controls access to resource identified by the key of the path object.
+
+```json
+"paths": {
+  "/me/activities/{id}": {
+    "leastPrivilegePath": ["DelegateedWork", "DelegatedPersonal"],
```
PetRich-MSFT, line 162:
```suggestion
    "leastPrivilegePath": ["DelegatedWork", "DelegatedPersonal"],
```

specs/permissions.md
```diff
@@ -0,0 +1,340 @@
...
+The "paths" member is a REQUIRED object whose keys contain a simplified URI template to identify the resources protected by this permission object.
+
+### alsoRequires
+The "alsoRequires" member is logical expression of permissions that must be presented as claims alongside the current permission. 
+
+```
+(User.Read | User.Read.All) & Group.Read
+```
+
+### includedProperties
```
PetRich-MSFT, line 112: May want to call out the behavior when **includeProperties** and **excludeProperties** are not specified. I assume **includeProperties** defaults to all properties if not specified, and **excludeProperties** defaults to no properties if not specified

specs/permissions.md
```diff
@@ -0,0 +1,340 @@
...
+] 
+```
+
+### schemeKeys
+The "schemeKeys" member is a REQUIRED array of strings that reference the schemes defined in the [permission object](#permissionObject) that are supported by the paths in this pathSet object. Each value in this array MUST match one of the keys of the "schemes" member in the [permission object](#permissionObject).
+
+### methods
+The "methods" member is a REQUIRED array of strings that represent the HTTP methods supported by the paths in this pathSet object.
+
+### paths
```
PetRich-MSFT, line 102: Should provide an example of how the URI template works

specs/permissions.md
```diff
@@ -0,0 +1,340 @@
...
+The "paths" member is a REQUIRED object whose keys contain a simplified URI template to identify the resources protected by this permission object.
```
PetRich-MSFT, line 103: How do we distinguish between **Read** and **Read.All** for paths permissions?

specs/permissions.md
```diff
@@ -0,0 +1,340 @@
...
+The "requiredEnvironments" member is an array of strings that identifies the deployment environments in which the permission SHOULD be supported. When this member is not present, support for all environments is implied.
+
+### privilegeLevel
+The "privilegeLevel" member value provides a hint as to the risks of consenting this permissions. Valid values include: low, medium and high.
+
+### appIdForConditionalAccessChecks
+TBD
+
+### ownerEmail
+The "ownerEmail" member is a REQUIRED string that provides a contact mechanism for communicating with the owner of the permission. It is important that owners of permissions are aware when new paths are added to an existing permission.
```
PetRich-MSFT, line 71: Suggestion: Require this to be a DL/Mail enabled security group

specs/permissions.md
```diff
@@ -0,0 +1,340 @@
...
+
+## <a name="pathSetObject"></a>PathSet Object
+A pathSet object identifies a set of paths that are accessible have a common set of security characteristics, such as HTTP methods and schemes. Ideally, a permission object contains a single pathSet object. This indicates that all paths protected by the permission support the same characteristics. In practice there are cases where support is not uniform. Distinct pathSet objects can be created to separate the paths with varying characteristics.
```
PetRich-MSFT, line 74:
```suggestion
A pathSet object identifies a set of paths that are accessible and have a common set of security characteristics, such as HTTP methods and schemes. Ideally, a permission object contains a single pathSet object. This indicates that all paths protected by the permission support the same characteristics. In practice there are cases where support is not uniform. Distinct pathSet objects can be created to separate the paths with varying characteristics.  
```

specs/permissions.md
```diff
@@ -0,0 +1,340 @@
...
+"paths": {
+  "/me/activities/{id}": {
+    "leastPrivilegePath": ["DelegateedWork", "DelegatedPersonal"],
+    "includedProperties": ["id","displayName"],
+    "excludedProperties": ["cost"]
+  }
+```
+
+### leastPrivilegePath
+The "leastPrivilegePath" member is an array of strings that identify the schemes for which this permission is the least privilege permission for accessing the path.
```
PetRich-MSFT, line 169: Question: can this not be derived automatically based on the set of paths and the **privilegeLevel** property?

specs/permissions.md
darrelmiller: Read and Read.All is a convention that Microsoft Graph has that indicates whether you can access any resource of the specified type, or just "yours".  It is a scoping mechanism that is opaque to this spec.  The model this spec describes currently does not have the capability to map permissions to caller identity.  At this time we are only trying to map claims to a URI template that we are primarily using to verify a user has consented an application to access data.  If you have any ideas how to describe user authZ as well, I am all ears.

specs/permissions.md
darrelmiller: That is what I originally thought and then someone convinced me that is not the case.  Let me try and remember what the reasoning was.  Nope, can't remember.  I will go back to the person and get them to remind me and I'll update this doc.  Thanks for the call out.

darrelmiller:
Add a provisioningInfo object. It will contain internal information:
owner  (description, _
deploymentEnvironments
resourceAppId
privileged  
hidden

Add Id into the main part of the spec.

specs/permissions.md
```diff
@@ -0,0 +1,349 @@
...
+
+The relationship between a permission and the resources it controls have a number of attributes including the HTTP method being granted, the type of security scheme being used, and properties of the resource being protected.  Ideally these relationships would be consistent for all resources controlled by a permission.  However, in reality that is not always the case. This model enables modelling the paths as sets of paths that do behave consistently and a permission can have multiple sets of paths. This approach minimizes the duplication of relationship attributes and encourages consistency in relationship design due to the overhead of managing distinct path sets. 
+
+## Notational Conventions
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in BCP 14 [RFC2119] [RFC8174] when, and only when, they appear in all capitals, as shown here.
```
wandeg, line 14: Should we have urls to the RFCs provided here?

PR #192, opened 2022-07-27
Schema sync 2022-07-27 from 1.2.13.1

marabooy:


marabooy:
Closing as there are new schemas that were released as part of 1.2.20.0

PR #210, opened 2022-10-05
Adds insertability and updatability for `administrativeUnit` navigation property

irvinesunday:
- `educationSchool/administrativeUnit` is a non-containment navigation property. We don't generate `Post` and `Patch/Update` operations for non-containment navigation properties unless explicitly set in restrictions annotations.


baywet:
@irvinesunday this PR is conflicting, would you mind updating it first please?

baywet (approved):

Issue #216, opened 2022-10-27
Replace `graph.report` return type with `Edm.Stream` return type only for report functions that start with `getXXX`


irvinesunday:
Related to https://github.com/microsoftgraph/msgraph-metadata/issues/140

This PR https://github.com/microsoftgraph/msgraph-metadata/pull/213 replaced `graph.report` return type with `Edm.Stream` return type for matching report functions. We want to tweak this to only update for report functions that start with `getXXX`. See rationale here: https://github.com/microsoftgraph/msgraph-sdk-java/pull/1245#issuecomment-1292048012

PR #217, opened 2022-10-27
Replaces `graph.report` return type with `Edm.Stream` return type only for report functions that start with `get`

irvinesunday:
Fixes https://github.com/microsoftgraph/msgraph-metadata/issues/216

This PR: 
- Replaces `graph.report` return type with `Edm.Stream` return type only for report functions that start with `get`.


baywet (approved):
thank you for turning around so quickly on this!

PR #221, opened 2022-11-04
Update permissions schema

wandeg:
- Update $schema entry to match the $id value of the draft-07 meta-schema. (Required for validation)
- Update schema to match the updates to the spec

### DIFF
--- a/graph-permissions-schema.json
+++ b/graph-permissions-schema.json
@@ -1,5 +1,5 @@
 {
-    "$schema": "https://json-schema.org/draft-07/schema#",
+    "$schema": "http://json-schema.org/draft-07/schema#",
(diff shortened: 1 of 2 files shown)


graph-permissions-schema.json
```diff
@@ -85,7 +85,12 @@
                 "schemeKeys": {
```
irvinesunday, line 85: Just out of curiosity... was the intention here to name this `schemeKeys` or `schemaKeys`? 🤔

wandeg: `schemeKeys` https://github.com/microsoftgraph/msgraph-metadata/blob/dm/permissionspec/specs/permissions.md#schemekeys

irvinesunday (approved):

Issue #237, opened 2022-12-21
restrictions annotations missing for /teams endpoint


baywet:
The service doesn't support the following call in v1/beta today:
- GET /teams

However, our current OpenAPI description is documenting this operation, leading clients to have an API surface for it, confusing customers.
To be clear, GET /teams/{id} [is supported](https://learn.microsoft.com/en-us/graph/api/team-get?view=graph-rest-1.0&tabs=http) and POST /teams [as well](https://learn.microsoft.com/en-us/graph/api/team-post?view=graph-rest-1.0&tabs=http)

We should add an annotation to the metadata XSL to this operation is not emitted anymore during the conversion process.

peombwa:
@baywet, the service support `GET /teams` in `beta` - https://learn.microsoft.com/en-us/graph/api/teams-list?view=graph-rest-beta&tabs=http. The restrictions annotations should only be applied to the `v1.0` metadata.

PR #239, opened 2022-12-22
Remove readability for the `teams` entity set

irvinesunday:
This PR:
- Fixes https://github.com/microsoftgraph/msgraph-metadata/issues/237
- Removes the `GET` operation for `/teams`
- Ensures that the `GET` operation for `/teams{team-id}` remains.


baywet (changes requested):
the test file needs to be updated

transforms/csdl/preprocess_csdl.xsl
```diff
@@ -926,6 +965,18 @@
...
        </xsl:copy>
     </xsl:template>
+
+    <!-- If the parent "Annotations" tag already exists modify it -->
+    <!-- Remove readability only for teams entity set -->
+    <xsl:template match="edm:Schema[@Namespace='microsoft.graph']/edm:Annotations[@Target='microsoft.graph.GraphService/teams']">
+      <xsl:copy>
+        <xsl:copy-of select="@*|node()"/>
+           <xsl:call-template name="ReadRestrictionsTemplate">
+              <xsl:with-param name="readable">false</xsl:with-param>
```
peombwa, line 975: We should isolate `readable` of `false` to just v1.0 metadata. `GET /teams` is [supported in `beta`](https://learn.microsoft.com/en-us/graph/api/teams-list?view=graph-rest-beta&tabs=http).

irvinesunday: Anyone know how we can be able to target the file path to the source XML? We might need to target the file path name to be able make this conditional decision. Trying [document-uri()](https://learning.oreilly.com/library/view/xslt-2nd-edition/9780596527211/re83.html) fails with the error: _document-uri()' is an unknown XSLT function._
@peombwa @baywet

baywet: we could imagine adding another parameter and updating the scripts if that makes things easier.

peombwa: +1 on using a parameter instead. This will make it easier to apply API version specific annotations in the future.

irvinesunday: All agreed. Will transfer this to be parameter-based

irvinesunday:
PR is temporarily blocked by: https://github.com/microsoftgraph/msgraph-metadata/issues/282

transforms/csdl/preprocess_csdl.xsl
irvinesunday: This has been resolved

transforms/csdl/preprocess_csdl.xsl
```diff
@@ -6,7 +6,7 @@
     <xsl:strip-space elements="*"/> <!-- Remove empty space after deletions. -->
     <xsl:param name="remove-capability-annotations">True</xsl:param>
     <xsl:param name="add-innererror-description">False</xsl:param>
-    <xsl:param name="csdlVersion">v1_0</xsl:param>
+    <xsl:param name="csdlVersion">v1.0</xsl:param>
```
baywet, line 9: May I suggest that we define custom functions so the comparison doesn't need to be done across the document?

https://stackoverflow.com/questions/16164205/create-custom-xslt-function

irvinesunday: Can you expound this further? How do you foresee using the `csdlVersion` param with custom functions? What are we aiming to resolve?

baywet: sure. With this param, every block that needs to use it needs to make the string comparison. Should the value change (like you pointed out from underscore to a period), every comparison will need to be updated. Whereas with a function, all the blocks can call the function (IsV1, IsBeta) and the comparison is only done at a single place.

irvinesunday: How about this approach?

https://github.com/microsoftgraph/msgraph-metadata/blob/1ac31d90adb4b3a99079ff422e222cdc2c69ff5b/transforms/csdl/preprocess_csdl.xsl#L20-L25

peombwa (approved):

Issue #238, opened 2022-12-21
`drive` nav property for `site` entityType should have `ContainsTarget="true"` attribute


peombwa:
The ODSP API supports uploading of files to a site's default document library via `PUT /sites/{site-id}/drive/items/{item-id}/content`. This path is currently missing in the OpenAPI document due to missing `ContainsTarget="true"` attribute on `drive` nav property in the clean metadata.

See https://learn.microsoft.com/en-us/graph/api/driveitem-put-content?view=graph-rest-1.0&tabs=http#http-request-to-replace-an-existing-item.

baywet:
I disagree, drive has a root collection of drives. Hence it's not contained.

PR #240, opened 2022-12-22
Add back containment for `microsoft.graph.site/drive` navigation property

irvinesunday:
This PR:
- Fixes https://github.com/microsoftgraph/msgraph-metadata/issues/238
- Removes the transform that removed `ContainsTarget=true` for the `drive` navigation property in the `site` entity type. This nav. property is now a containment nav. prop.

**NB:** Adding back the containment of this nav. prop. bumps up our OpenAPI document size from **7,951** to **8,317** paths in v1.0. Is this something that we need to assess further before proceeding with the merge?


baywet (changes requested):
the test file needs to be updated

irvinesunday:
> the test file needs to be updated

There is no side effect in the _preprocess_csdl_test_output.xml_ file with this change.

baywet:
Are you able to run ./scripts/run-metadata-validation.ps1  locally? what's the memory use of the script?

baywet:
we shouldn't implement that change as it'd cause expansion issues.

Issue #84, opened 2021-08-18
[Breaking Change] Revert deleting workbooks entity set


irvinesunday:
This [update](https://github.com/microsoftgraph/msgraph-metadata/commit/8e97b2f5bd7358b195556fb4cdc4a4191d85f5f6#diff-3964e1d46914c457caec43f8f7aa1430a486f0a78419e24ec21e7f4e08cdf00e) to the metadata that happened on 20th July 2021 removed the workbooks entity set. This caused the generation of multiple paths due to the expansion of navigation property paths of type `driveItem` during conversion of the CSDL to OpenAPI by the `Microsoft.OpenApi.OData` lib. `v1.0.8` used by the DevX API. This lead to service unavailability in the API due to memory over-utilisation.
The `workbooks` entity set needs to be reinstated. 
[AB#10770](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/10770)

irvinesunday:
Re-opening this issue as a temporary fix to unblock PowerShell.

PR #247, opened 2023-01-16
- removes ghost workbooks root path

baywet:
related #86 #84. Adding the workbooks root node was already contreversial at the time. It's leading people to confusion and causing issues with over expansion on all OpenAPI based generation. @peombwa I understand this might be problematic for PS somehow, would you mind providing more context here? I'd support shipping a breaking change fix to remove something that never worked rather than maintaining the illusion it does work.

### DIFF
--- a/transforms/csdl/preprocess_csdl.xsl
+++ b/transforms/csdl/preprocess_csdl.xsl
@@ -941,20 +941,6 @@
     <!-- Remove directoryObject Capability Annotations -->
     <xsl:template match="edm:Schema[starts-with(@Namespace, 'microsoft.graph')]/edm:Annotations[@Target='microsoft.graph.directoryObject']/*[starts-with(@Term, 'Org.OData.Capabilities')]"/>
 
-    <!-- Add workbooks entity set if missing -->
-    <xsl:template match="edm:Schema[@Namespace='microsoft.graph']/edm:EntityContainer[@Name='GraphService']">
-        <xsl:copy>
-            <xsl:apply-templates select="@*"/>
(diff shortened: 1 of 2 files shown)


baywet:
Thanks for reverting Peter, we also went through memory issues until we fixed inaccuracies in the metadata as describes in #248. Can you have a look and let us know what you think of the findings please?

irvinesunday:
> @baywet, we don't expose commands that call workbooks in PowerShell.
> 
> I believe removing workbooks led to service unavailability in DevX API due to memory over-utilization:
> 
> > This caused the generation of multiple paths due to the expansion of navigation property paths of type driveItem during conversion of the CSDL to OpenAPI by the Microsoft.OpenApi.OData lib. v1.0.8 used by the DevX API. This lead to service unavailability in the API due to memory over-utilization.
> 
> It is the outage of DevX API that affected generation of PowerShell SDK. @irvinesunday, is my understanding of #84 correct?

This was before with the old version (< `v1.0.8`) of the conversion lib that used to check whether the entity type of the navigation property was defined as an entity type of any of the declared navigation sources to determine whether or not to expand the said navigation property. In later versions we use the attribute `ContainsTarget=true`, hence we no longer need to maintain the `workbooks` entity set.

irvinesunday (approved):

peombwa (approved):

baywet:
Thanks for pointing to the version difference. It allowed me to notice the validation tool here was behind. I fixed the dependabot configuration and merged the updates :)

PR #257, opened 2023-01-26
- replaces custom validation tools by hidi

### LINKED ISSUE #174: Replace OpenAPIParser validation with hidi
Why not use hidi with the validate command instead?

_Originally posted by @baywet in https://github.com/microsoftgraph/msgraph-metadata/pull/173#discussion_r908353064_

baywet:
- fixes #174

Signed-off-by: Vincent Biret <vibiret@microsoft.com>


baywet:
Impacted by both
https://github.com/microsoft/OpenAPI.NET/issues/1141
https://github.com/microsoft/OpenAPI.NET/issues/1142

baywet:
@MIchaelMainer I've put together this PR to align the validation experience with the one we have in the weekly generation pipeline and reduce maintenance here. Since we've merged @255 the validation keeps timing out (before this change). I believe that at this point the only difference with the weekly generation is the size of the agent.
Do you happen to know if we have [larger runners](https://docs.github.com/en/actions/using-github-hosted-runners/using-larger-runners) configured and if so, what are the group names? If not, can you create one at least for troubleshooting purposes at this point please?

baywet:
Actually interestingly the weekly generation is looking at the clean metadata instead of the raw one. https://github.com/microsoftgraph/MSGraph-SDK-Code-Generator/blob/bc4cf7eaba6c8acfa91b4a73f4a3e50547f1fc8d/scripts/generate-open-api.ps1#L18

So maybe the solution here is to validate the clean metadata instead? Adding @zengin to the rescue since I believe he setup most of the generation and validation pipeline and we probably want to avoid duplication.

zengin:
@baywet I agree. We should validate only the transformed metadata.

The idea behind the custom validation here was to make sure that we don't break downstream, and we catch simple parsing errors early at the time of PR. Looks like running `hidi` here is a fair and welcome replacement.

zengin (approved):

zengin:
side note: I had a GH issue #174 on this. Marked it in the PR description.

baywet:
Thanks for jumping in, and for tagging the issue, I had forgotten about it.
So should I update the paths of the csdl that's being used for validation at this point and merge?

zengin:
> So should I update the paths of the csdl that's being used for validation at this point and merge?

We are getting the transformed metadata by applying the transform in the check. Removing those steps will miss the errors that may be introduced by the transformation rules a PR may add.

baywet:
There's probably something that the generation pipeline is doing that we're not doing here during the validation. And hidi ends up expanding forever. Would you mind having a look? I think I'm missing something.

zengin:
I will take a look.

zengin:
@baywet, it turns out that there is a different transformation rule applied in the OpenAPI flow in the generation pipeline. Once we gave the expected parameters, `hidi` transform worked as expected.

https://github.com/microsoftgraph/msgraph-metadata/pull/257/commits/89c2621e26578445301e2973d75a7ef424e348a1

baywet:
Thanks for looking into this! I knew I was missing something but couldn't figure out what.

Issue #262, opened 2023-02-02
Conflicting searchQuery query string properties


baywet:
Currently beta contains the following complex type

```xml
<ComplexType Name="searchQuery">
        <Property Name="queryString" Type="Edm.String">
          <Annotation Term="Org.OData.Core.V1.Description" String="The search query containing the search terms. Required." />
        </Property>
        <Property Name="queryTemplate" Type="Edm.String">
          <Annotation Term="Org.OData.Core.V1.Description" String="Provides a way to decorate the query string. Supports both KQL and query variables. Optional." />
        </Property>
        <Property Name="query_string" Type="graph.searchQueryString" />
      </ComplexType>
```

This is causing conflicts across the board because of the two query string properties. If they are being mangled (which they are in most languages) they end up colliding.

We should remove the "query_string" property since it doesn't follow graph naming conventions.

PR #266, opened 2023-02-06
Removes invalid property `query_string` from complex property `searchQuery`

irvinesunday:
Fixes https://github.com/microsoftgraph/msgraph-metadata/issues/262

This PR:
- Removes the property `query_string` from the ComplexProperty `searchQuery`
- Updates the integration test files.


baywet (approved):
Thanks for handling it so fast!

Issue #270, opened 2023-02-10
Add back `NavigationRestrictions` that removes navigability for `driveItem/workbook` navigation property


irvinesunday:
This [PR](https://github.com/microsoftgraph/msgraph-metadata/pull/255) removed the `NavigationRestrictions` that removed the navigability for the `driveItem/workbook` navigation property, which meant further expanding the navigation property `workbook/worksheets` whose navigation properties are all contained (for the _cleanMetadataWithDescriptionsAndAnnotations_ CSDL file). This inadvertently affected the generation of OpenAPI files relying on the the above file (v1.0 and beta) by over-expanding the workbook paths and bloating the OpenAPI document, on top of causing a timeout in the OpenAPI generation process. We need to add back this restriction annotation to return the OpenAPI generation from the aforementioned CSDL file to its previous good state.

PR #271, opened 2023-02-10
Adds back `NavigationRestrictions` that removes navigability for `driveItem/workbook` navigation property

irvinesunday:
Fixes https://github.com/microsoftgraph/msgraph-metadata/issues/270

This PR:
- Removes the navigability for `driveItem/workbook` navigation property for the target CSDL file - `cleanMetadataWithDescriptionsAndAnnotations` (v1.0 and beta)

### DIFF
--- a/transforms/csdl/preprocess_csdl.xsl
+++ b/transforms/csdl/preprocess_csdl.xsl
@@ -7,7 +7,7 @@
     <xsl:param name="remove-capability-annotations">True</xsl:param>
     <xsl:param name="add-innererror-description">False</xsl:param>
 
-    <!-- Flag to signal if we are generating a document for open api generation. -->
+    <!-- Flag to signal if we are generating a document for Kiota-based open api generation. -->
     <xsl:variable name="open-api-generation">
         <xsl:choose>
             <!-- Open API document generation is done with capability annotations and error descriptions -->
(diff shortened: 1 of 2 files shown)


transforms/csdl/preprocess_csdl.xsl
```diff
@@ -786,6 +786,25 @@
     <xsl:template match="edm:Schema[@Namespace='microsoft.graph']">
         <xsl:copy>
             <xsl:apply-templates select="@* | node()"/>
+            
+            <!-- Remove navigability for driveItem/workbook navigation property for DevX API-specific CSDL-->
+            <xsl:choose>
+                <xsl:when test="$remove-capability-annotations='False'">
```
baywet, line 792: assuming this is true for kiota based SDKs?

andrueastman: OpenApi generation for Kiota based SDKs does include annotations so this navigability will also be removed for Kiota based SDKs. 
To omit the file for Kiota based SDKs we may need to include the `open-api-generation` variable in the test [here ](https://github.com/microsoftgraph/msgraph-metadata/blob/master/transforms/csdl/preprocess_csdl.xsl#L11)

baywet: Thanks for the context. I believe kiota deceptions should not have the restriction. Irvine, can you make the change please?

irvinesunday: This is effected.

baywet (approved):

Issue #276, opened 2023-02-21
beta generation failing because of delta function


baywet:
[beta weekly generation is currently failing](https://microsoftgraph.visualstudio.com/Graph%20Developer%20Experiences/_build/results?buildId=106293&view=logs&j=4c81e8e4-6eb9-5def-7261-b44ad0fc1b7d&t=163839c6-f69e-5b5d-39e2-5b38e97d3fe3) because [a delta function on directory objects was added](https://dev.azure.com/msazure/One/_git/AD-AggregatorService-Workloads/pullRequest/7582888#1676991425) which conflicts with the derived types delta functions.

I believe we should introduce an xsl fix that removes the functions on the derived types, only for beta, only for kiota based SDK for the time being until the service fixes their definition to unblock our generation pipeline, but I'm open to feedback here first.

CC @andrueastman @peombwa @irvinesunday

peombwa:
+1 to adding an XSL transform to remove the functions on the derived types (I assume OpenAPI.NET.OData will still add the function to the derived types accordingly). This XSL transform should be applied to all metadata files used by OpenAPI tools (Kiota, DevX, etc.).

andrueastman:
> +1 to adding an XSL transform to remove the functions on the derived types (I assume OpenAPI.NET.OData will still add the function to the derived types accordingly). This XSL transform should be applied to all metadata files used by OpenAPI tools (Kiota, DevX, etc.).

I agree. Implement the transform on the metadata but still eventually make the conversion library resilient to this as this seems to be valid OData according to https://github.com/microsoft/OpenAPI.NET.OData/issues/343#issuecomment-1438940862

irvinesunday:
This is currently affecting `beta`, do we want to be proactive and cover `v1` as well?

baywet:
just beta for the time being since the change hasn't rolled out to v1 at the moment. I suspect it'll eventually come to v1, and we'll update/remove the condition when it does.

andrueastman:
Looks like this one was recently merged(yesterday) so we will need it for v1 as well. 
https://msazure.visualstudio.com/One/_git/AD-AggregatorService-Workloads/pullrequest/7644692

baywet:
closing since the fix is implemented in the conversion lib

PR #279, opened 2023-02-22
Removes delta function for the derived type `administrativeUnit`

irvinesunday:
Fixes https://github.com/microsoftgraph/msgraph-metadata/issues/276

This PR:
- Removes the `delta` function of type `Collection(graph.administrativeUnit)`.
- Adds a test to validate this.

### DIFF
--- a/transforms/csdl/preprocess_csdl.xsl
+++ b/transforms/csdl/preprocess_csdl.xsl
@@ -435,7 +435,7 @@
     <!--Remove functions that are blocking beta generation-->
     <xsl:template match="edm:Schema[@Namespace='microsoft.graph']/edm:Function[@Name='delta'][edm:Parameter[@Name='token']][edm:Parameter[@Type='Collection(graph.site)']]"/>
     <xsl:template match="edm:Schema[@Namespace='microsoft.graph']/edm:Function[@Name='additionalAccess'][edm:Parameter[@Name='accessPackageId']][edm:Parameter[@Type='Collection(graph.accessPackageAssignment)']][1]"/>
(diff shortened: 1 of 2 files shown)


baywet (changes requested):
I think that once we remove it for that type, the next type (alphabetically) will fail too, we probably need to remove it for all types that derive from directory object (either directly or indirectly), what do you think?

irvinesunday:
> I think that once we remove it for that type, the next type (alphabetically) will fail too, we probably need to remove it for all types that derive from directory object (either directly or indirectly), what do you think?

Since the base type probably has several derived types, I think it will be better having the duplicate path generation issue fixed in the conversion lib. Otherwise, it will require identifying all the derived types and applying their fixes.

baywet:
> Since the base type probably has several derived types, I think it will be better having the duplicate path generation issue fixed in the conversion lib. Otherwise, it will require identifying all the derived types and applying their fixes.

Agreed that it'll scale better and ease up the maintenance. The question the generation is now blocked. How soon can we get the fix out to unblock it if we want to skip the XSLT fix?

irvinesunday:
> > Since the base type probably has several derived types, I think it will be better having the duplicate path generation issue fixed in the conversion lib. Otherwise, it will require identifying all the derived types and applying their fixes.
> 
> Agreed that it'll scale better and ease up the maintenance. The question the generation is now blocked. How soon can we get the fix out to unblock it if we want to skip the XSLT fix?

I'll work on it ASAP.

baywet:
Closing since the fix is implemented in the conversion lib now.

PR #284, opened 2023-02-22
Task: Add optional parameter to support validation

thewahome:
Depends on the merging of [this PR](https://github.com/microsoftgraph/MSGraph-SDK-Code-Generator/pull/944) .

The new generator will need to validate the individual "platform" yaml files generated. It passes a new parameter "platformName" that tells the validation script which file to validate.

### Changes
This script will change from:
> ./scripts/run-openapi-validation.ps1 -repoDirectory C:/github/msgraph-metadata -version "v1.0" 

To
> ./scripts/run-openapi-validation.ps1 -repoDirectory C:/github/msgraph-metadata -version "v1.0" -platformName "openapi"

### Notes
To make sure things do not break, the `platformName` parameter is optional and set to the known available platform -> `openapi` to make sure things do not break

### DIFF
--- a/.github/workflows/metadata-parser-validation.yml
+++ b/.github/workflows/metadata-parser-validation.yml
@@ -14,7 +14,7 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - id: set-conversion-settings
-        run: echo "::set-output name=matrix::$(ls conversion-settings/*.json | jq -R -f -s -c 'split("\n")[:-1]')"
+        run: echo "::set-output name=matrix::$(ls conversion-settings/*.json | sed 's/.*\///' | sed 's/\.[^.]*$//' | jq -R -s -c 'split("\n")[:-1]')"
 
   build:
     needs: list-conversion-settings
--- a/.github/workflows/metadata-parser-validation.yml
+++ b/.github/workflows/metadata-parser-validation.yml
@@ -40,6 +40,7 @@ jobs:
       shell: pwsh
 
     - uses: actions/upload-artifact@v3
+      if: always()
       with:
         name: openapi
         path: openapi/*
--- a/.github/workflows/metadata-parser-validation.yml
+++ b/.github/workflows/metadata-parser-validation.yml
@@ -7,10 +7,21 @@ on:
     branches: [ master ]
 
 jobs:
+  list-conversion-settings:
+    runs-on: ubuntu-latest
+    outputs:
+      matrix: ${{ steps.set-conversion-settings.outputs.matrix }}
+    steps:
+      - uses: actions/checkout@v2
+      - id: set-conversion-settings
+        run: echo "::set-output name=matrix::$(ls conversion-settings/*.json | jq -R -f -s -c 'split("\n")[:-1]')"
+
   build:
+    needs: list-conversion-settings
     strategy:
       matrix:
         version: ["v1.0", "beta"]
+        settings: ${{ fromJson(needs.list-conversion-settings.outputs.matrix) }}
 
     runs-on: ubuntu-latest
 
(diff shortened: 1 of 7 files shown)


scripts/run-openapi-validation.ps1
```diff
@@ -16,15 +16,25 @@
...
 
 param(
     [Parameter(Mandatory=$true)][string]$repoDirectory,
-    [Parameter(Mandatory=$true)][string]$version
+    [Parameter(Mandatory=$true)][string]$version,
+    [Parameter(Mandatory=$false)][string]$platformName
 )
-$yaml = Join-Path $repoDirectory "openapi" $version "openapi.yaml"
+
+if([string]::IsNullOrWhiteSpace($aString))
```
baywet, line 30: where is aString supposed to come from?

baywet (changes requested):
we probably need the convert settings for that step as well.
https://github.com/microsoftgraph/msgraph-metadata/blob/c67be8e3838be1d26e474fe060d269646833930d/scripts/run-metadata-validation.ps1#L39

scripts/run-openapi-validation.ps1
thewahome: Lemme fix that. It's supposed to be platformName

thewahome:
> we probably need the convert settings for that step as well.
> https://github.com/microsoftgraph/msgraph-metadata/blob/c67be8e3838be1d26e474fe060d269646833930d/scripts/run-metadata-validation.ps1#L39

I probably have low context on what this file does however it wasn't run by the pipeline step that generates openapi files hence it's not required to be specific to any platform

thewahome:
> we probably need the convert settings for that step as well.
> https://github.com/microsoftgraph/msgraph-metadata/blob/c67be8e3838be1d26e474fe060d269646833930d/scripts/run-metadata-validation.ps1#L39

I probably have low context on what this file does however it wasn't run by the pipeline step that generates openapi files hence it's not required to be specific to any platform

baywet:
> I probably have low context on what this file does however it wasn't run by the pipeline step that generates openapi files hence it's not required to be specific to any platform

It validates changes made to the XSL /transformation rules before they are committed to the repo. That's to ensure any change to the XSL is working for the CSDL and not breaking the OpenAPI conversion. And since we're planning on changing the defaults in hidi, that validation will be impacted and should load settings. Ideally it'd actually validate against all settings using a matrix to make sure we smoke test for all of our experiences.

thewahome:
>  And since we're planning on changing the defaults in hidi, that validation will be impacted and should load settings. Ideally it'd actually validate against all settings using a matrix to make sure we smoke test for all of our experiences.

@baywet there are no longer any changes needed on Hidi. I am adding a matrix to smoke test this

scripts/run-openapi-validation.ps1
baywet:
```suggestion
if([string]::IsNullOrWhiteSpace($platformName))
```

.github/workflows/metadata-parser-validation.yml
```diff
@@ -7,21 +7,39 @@ on:
...
       uses: actions/setup-dotnet@v1
       with:
         dotnet-version: 7.0.x
+
     - name: Validate latest XSLT rules
-      run: ./scripts/run-metadata-validation.ps1 -repoDirectory "${{ github.workspace }}" -version "${{ matrix.version }}"
+      run: ./scripts/run-metadata-validation.ps1 -repoDirectory "${{ github.workspace }}" -version "${{ matrix.version }}" -platformName "${{ matrix.settings }}"
       shell: pwsh
+
+    - uses: actions/upload-artifact@v3
```
baywet, line 42: very good initiative to upload things as artifacts!
```suggestion
    - uses: actions/upload-artifact@v3
      if: always()
```

baywet (changes requested):
Thanks for moving things around. couple of minor comments

baywet:
@thewahome can you double check the matrix aspect, it doesn't seem to be triggering the build jobs anymore. Thanks!

thewahome:
> @thewahome can you double check the matrix aspect, it doesn't seem to be triggering the build jobs anymore. Thanks!

On it @baywet

thewahome:
@baywet this is done.
The validations should fail until when the [SDK generator](https://github.com/microsoftgraph/MSGraph-SDK-Code-Generator/) is triggered to create the files

baywet:
@thewahome could we add placeholders by just copy/pasting the existing ones just so we can make sure the pipeline changes work?

thewahome:
> @thewahome could we add placeholders by just copy/pasting the existing ones just so we can make sure the pipeline changes work?

The key things to check are:
- whether the pipeline spawns the combinations of version and platform validation jobs.
- whether the spawned job checks in the correct folder location where the expected files will be placed.

We can see the spawned jobs. And on checking the one that failed we can see the error message shows that it went looking in the right place

> Could not validate the document, reason: 
> Could not open the file at /home/runner/work/msgraph-metadata/msgraph-metadata/openapi/v1.0/graphexplorer.yaml

baywet (approved):

PR #291, opened 2023-03-02
Task: adds default settings

thewahome:
The generated open api document should be used by power platforms and the plain devx api style setting

### DIFF
--- a/conversion-settings/default.json
+++ b/conversion-settings/default.json
@@ -1,0 +1,24 @@
+{
+  "OpenApiConvertSettings": {
+    "AddEnumDescriptionExtension": true,
+    "AddSingleQuotesForStringParameters": true,
+    "DeclarePathParametersOnPathItem": true,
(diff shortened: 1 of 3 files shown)


baywet (approved):

irvinesunday (changes requested):
I believe these should be the default settings:

```CSharp
AddSingleQuotesForStringParameters = true,
AddEnumDescriptionExtension = true,
EnableKeyAsSegment = true,
EnableOperationId = true,
PrefixEntityTypeNameBeforeKey = true,
TagDepth = 2,
EnablePagination = true,
EnableDiscriminatorValue = false,
EnableDerivedTypesReferencesForRequestBody = false,
EnableDerivedTypesReferencesForResponses = false,
ExpandDerivedTypesNavigationProperties = false
```

irvinesunday (approved):

PR #319, opened 2023-04-12
Add custom query options for includeHiddenFolders and includeHiddenMessages

andrueastman:
Closes https://github.com/microsoftgraph/msgraph-sdk-dotnet/issues/1814
Closes https://github.com/microsoftgraph/msgraph-sdk-dotnet/issues/1680

Adds a `CustomQueryOptionsTemplate` to enable adding custom query option to various navigation properties as needed and adds `includeHiddenFolders` and  `includeHiddenMessages` query options for relevant endpoints

Ref - https://learn.microsoft.com/en-us/graph/api/user-list-mailfolders?view=graph-rest-1.0&tabs=csharp

### DIFF
--- a/transforms/csdl/preprocess_csdl.xsl
+++ b/transforms/csdl/preprocess_csdl.xsl
@@ -531,6 +531,42 @@
         </xsl:element>
     </xsl:template>
 
+    <!-- Add custom query options to given navigation property -->
+    <xsl:template name="CustomQueryOptionsTemplate">
+        <xsl:param name = "customQueryOptionName" />
+        <xsl:param name = "description" />
+        <xsl:element name="Record">
+            <xsl:element name="PropertyValue">
+                <xsl:attribute name="Property">ReadRestrictions</xsl:attribute>
+                <xsl:element name="Record">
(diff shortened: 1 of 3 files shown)


baywet (commented):
Should we do it for the group as well?

andrueastman:
@baywet Just to confirm, do you mean the annotations to `mailFolders` and `messages` in a group? According to the docs and the metadata, there currently are no such paths in a group. These accessing these collections are tied to the user.
https://learn.microsoft.com/en-us/graph/api/user-list-messages?view=graph-rest-1.0&tabs=http
https://learn.microsoft.com/en-us/graph/api/user-list-mailfolders?view=graph-rest-1.0&tabs=csharp

baywet:
Yeah that's what I meant but group messages are conversations. Even though under the hood they are technically the same.
https://learn.microsoft.com/en-us/graph/api/resources/group?view=graph-rest-1.0

MartinM85:
@baywet @andrueastman Could you please explain what are hidden messages? There is no documentation for includeHiddenMessages query parameter.

PR #343, opened 2023-05-09
athmangude/permissionspec

athmangude:
### Changes

- Moved `alsoRequires` and `implicit` properties into path object from the permissions object
- Added usage for `alsoRequires` for simple and complex expressions of additional permissions
- Changed the value of `path` object to string
- Update model diagram

### DIFF
--- a/specs/permissions.md
+++ b/specs/permissions.md
@@ -61,7 +61,7 @@ The "pathSets" member is a REQUIRED JSON Array. Each element of the array is a [
 
 The owner info object contains information related to the ownership of the permission. This object should only contain information that is not required by a consumer of the API and can safely be removed in any public projection of the permissions information.
 ### ownerSecurityGroup
-The "ownerSecurityGroup" member is a string that provides a contact mechanism for communicating with the owners of the permission. It is important that owners of permissions are aware when new paths are added to an existing permission. This member is only required when `ownerInfo` is present.
+The "ownerSecurityGroup" member is a string that provides a contact mechanism for communicating with the owners of the permission. It is important that owners of permissions are aware when new paths are added to an existing permission. This member is REQUIRED when `ownerInfo` is present.
 
 ## <a name="pathSetObject"></a>PathSet Object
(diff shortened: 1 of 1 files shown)


specs/permissions.md
```diff
@@ -137,15 +115,18 @@ The scheme object has members that describe the permission within the context of
         "adminDescription": "Allows the app to read and report the signed-in user's activity in the app.",
         "userConsentDisplayName": "Read and write app activity to users'activity feed",
         "userConsentDescription": "Allows the app to read and report the signed-in user's activity in the app.",
-        "requiresAdminConsent": true
+        "requiresAdminConsent": true,
+        "privilegeLevel": 3
```
darrelmiller, line 119: Using integer values is problematic because how does anyone know if 1 is high privilege or 5 is?

specs/permissions.md
```diff
@@ -51,33 +51,15 @@ The permissions object contains the details about a permission that can be used
...
-### requiredEnvironments
-The "requiredEnvironments" member is an array of strings that identifies the deployment environments in which the permission SHOULD be supported. When this member is not present, support for all environments is implied.
-
-### resourceAppId
-The "resourceAppId" member value provides an identifier of the resource server that is used to enforce Conditional Access checks for this permission.
+## <a name="ownerInfo"></a>Owner Info Object
 
+The owner info object contains information related to the ownership of the permission. This object should only contain information that is not required by a consumer of the API and can safely be removed in any public projection of the permissions information.
 ### ownerSecurityGroup
 The "ownerSecurityGroup" member is a REQUIRED string that provides a contact mechanism for communicating with the owners of the permission. It is important that owners of permissions are aware when new paths are added to an existing permission.
```
darrelmiller, line 64: We should qualify that ownerSecurityGroup is only required when ownerInfo is present otherwise it is a contradiction saying that it can be "safely removed".

specs/permissions.md
```diff
@@ -113,9 +95,6 @@ The "methods" member is a REQUIRED array of strings that represent the HTTP meth
 ### paths
 The "paths" member is a REQUIRED object whose keys contain a simplified URI template to identify the resources protected by this permission object.
 
-### alsoRequires
-The "alsoRequires" member is logical expression of permissions that must be presented as claims alongside the current permission. 
-
```
darrelmiller, line 118: This needs to go back here.

Issue #305, opened 2023-03-24
Consider setting AddAlternateKeyPaths to true in the conversion settings


peombwa:
APIs such as [List members of a directory role](https://learn.microsoft.com/en-us/graph/api/directoryrole-list-members?view=graph-rest-1.0) now support [alternate keys](https://docs.oasis-open.org/odata/odata/v4.01/os/part2-url-conventions/odata-v4.01-os-part2-url-conventions.html#sec_AlternateKeys):
``` http
GET /directoryRoles/{role-id}/members
GET /directoryRoles(roleTemplateId='{roleTemplateId}')/members
```

We should consider setting `AddAlternateKeyPaths` to `true` in the CSDL to OpenAPI conversion settings at https://github.com/microsoftgraph/msgraph-metadata/tree/master/conversion-settings to support the alternate key in our SDKs.

As part of this change, we should monitor the net effect of this setting on the growth of the number of paths.

irvinesunday:
I will do a path count diff of the two files with the setting `true` and `false` and share the results here.

darrelmiller:
/cc @baywet Does Kiota currently handle this kind of alternate path syntax?

@irvinesunday  I think this issue should be transferred over to the OpenAPI.OData  repo.

baywet:
@darrelmiller yes, you'll get a directoryRolesByRoleTemplateId method in the fluent API surface.
Since we [already have the setting in the conversion lib](https://github.com/microsoft/OpenAPI.NET.OData/blob/8fcb0461050d21d5b133a97a5c454f94c5723296/src/Microsoft.OpenApi.OData.Reader/OpenApiConvertSettings.cs#L155) is merely a matter of updating the configuration files here, no action is needed in the conversion lib unless I missed something. So I'd argue against moving this issue.

njaci1:
@irvinesunday how did this go?

irvinesunday:
Setting `AddAlternateKeyPaths` increases the paths for v1.0 from 6237 - 6241 paths. This is quite an insignificant increase.

PR #362, opened 2023-06-15
Adds `AddAlternateKeyPaths = true`

irvinesunday:
Fixes https://github.com/microsoftgraph/msgraph-metadata/issues/305

### DIFF
--- a/conversion-settings/openapi.json
+++ b/conversion-settings/openapi.json
@@ -19,6 +19,7 @@
     "TagDepth": 2,
     "UseSuccessStatusCodeRange": true,
     "NamespacePrefixToStripForInMethodPaths": "microsoft.graph",
-    "EnableAliasForTypeCastSegments": true
+    "EnableAliasForTypeCastSegments": true,
+    "AddAlternateKeyPaths": true
   }
 }
--- a/conversion-settings/powershell.json
+++ b/conversion-settings/powershell.json
@@ -15,6 +15,7 @@
     "ShowLinks": false,
     "DeclarePathParametersOnPathItem": false,
     "EnableODataAnnotationReferencesForResponses": false,
-    "EnableDollarCountPath": false
+    "EnableDollarCountPath": false,
+    "AddAlternateKeyPaths": true
   }
 }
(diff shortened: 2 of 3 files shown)


andrueastman (commented):
Should we also do this for the other setting files as well?

irvinesunday:
> Should we also do this for the other setting files as well?

If we enable this setting in the OpenAPI file used by GE we might not extract accurate path segments for these alternate key paths in the OpenApiUrlTreeNode which is used by Resource Explorer.
We split the path value by `/` when creating the tree node: https://github.com/microsoft/OpenAPI.NET/blob/9d82233182d11d64a12bd5c250b881e6e102f4d0/src/Microsoft.OpenApi/Services/OpenApiUrlTreeNode.cs#L158

This should be resolved in the OpenAPI.NET library first. I have created an issue to track this: https://github.com/microsoft/OpenAPI.NET/issues/1277

baywet:
I think I agree with @andrueastman and we should at least set it for the openapi profile. Left a comment on the issue you created to understand the concern better.

irvinesunday:
> I think I agree with @andrueastman and we should at least set it for the openapi profile. Left a comment on the issue you created to understand the concern better.

@andrueastman is suggesting we implement this change for all the profiles. My suggestion was to localise it just to **powershell** for the time being. Does Kiota also anticipate the alternate key paths in its _openapi_ document?

baywet:
looking at [this comment](https://github.com/microsoft/OpenAPI.NET/issues/1277#issuecomment-1621806333) kiota would give you:
- directoryRoles/{id} => ...DirectoryRoles["id"] or ...DirectoryRoles.ById("id")
- directoryRoles(roleTemplateId='{roleTemplateId}') => DirectoryRolesByRoleTemplateId("roleTemplateId") (or With instead of By, I can't remember which convention we ended up landing on)

andrueastman:
> directoryRoles(roleTemplateId='{roleTemplateId}') => DirectoryRolesByRoleTemplateId("roleTemplateId") (or With instead of By, I can't remember which convention we ended up landing on)

It would be `DirectoryRolesWithRoleTemplateId`. 

As @baywet highlights, Kiota should be able to handle both paths without issue and having them in the openApi reference would unlock the alternate paths in the SDK request builders.

baywet (approved):

PR #295, opened 2023-03-08
Adds insertability for drive/bundles

andrueastman:
Related to https://github.com/microsoftgraph/msgraph-sdk-dotnet/issues/1630

Adds insert ability  `drive/bundles` so that we we can have Post method for creating bundles in Kiota based SDKs.

Related documentation.
https://learn.microsoft.com/en-us/graph/api/drive-post-bundles?view=graph-rest-1.0&tabs=http

PR #368, opened 2023-06-23
Add Insertability for adminUnits/members property

ramsessanchez:
AdminUnits/members should be insertable according to the documentation. 
https://learn.microsoft.com/en-us/graph/api/administrativeunit-post-members?view=graph-rest-1.0&tabs=java#http-request


andrueastman (commented):
Thanks for this @ramsessanchez 
I believe we just need an additional statement in case the annotation already exists (or is added) and needs to be modified. 

Checkout this PR for reference
https://github.com/microsoftgraph/msgraph-metadata/pull/295/files

andrueastman (approved):

Issue #329, opened 2023-04-19
Add the query parameters to the events delta functions


andrueastman:
Related to https://github.com/microsoftgraph/msgraph-sdk-go/issues/318

rsitrifork:
Is there any update on when this might be fixed, or any workaround? The Start and EndDate parameters are required for this API, so it is unusable without being able to set them.

Is there any documentation on how we might be able to contribute the changes to the schema ourselves?

PR #369, opened 2023-06-26
Fix event delta function parameters to include start and end dates

andrueastman:
Closes https://github.com/microsoftgraph/msgraph-metadata/issues/329
Closes https://github.com/microsoftgraph/msgraph-sdk-go/issues/318
Closes https://github.com/microsoftgraph/msgraph-sdk-dotnet/issues/1860
Closes https://github.com/microsoftgraph/msgraph-sdk-php/issues/1152

Uses the `Org.OData.Capabilities.V1.OperationRestrictions` to add the custom query parameters to the odata function.

### DIFF
--- a/transforms/csdl/preprocess_csdl.xsl
+++ b/transforms/csdl/preprocess_csdl.xsl
@@ -487,6 +487,31 @@
         <xsl:attribute name="Type">Collection(graph.conditionalAccessConditions)</xsl:attribute>
     </xsl:template>
     
+    <!--Delta function for events need the start and end date parameters-->
+    <xsl:template match="edm:Schema[@Namespace='microsoft.graph']/edm:Function[@Name='delta'][edm:Parameter[@Name='bindingparameter']][edm:Parameter[@Type='Collection(graph.event)']]">
+        <xsl:copy>
+            <xsl:apply-templates select="@* | node()"/>
+            <Annotation Term="Org.OData.Capabilities.V1.OperationRestrictions">
+            <Record>
+                <PropertyValue Property="CustomQueryOptions">
+                    <Collection>
+                        <Record>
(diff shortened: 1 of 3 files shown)


transforms/csdl/preprocess_csdl.xsl
```diff
@@ -487,6 +487,31 @@
...
+    <!--Delta function for events need the start and end date parameters-->
+    <xsl:template match="edm:Schema[@Namespace='microsoft.graph']/edm:Function[@Name='delta'][edm:Parameter[@Name='bindingparameter']][edm:Parameter[@Type='Collection(graph.event)']]">
+        <xsl:copy>
+            <xsl:apply-templates select="@* | node()"/>
+            <Annotation Term="Org.OData.Capabilities.V1.OperationRestrictions">
+            <Record>
+                <PropertyValue Property="CustomQueryOptions">
+                    <Collection>
+                        <Record>
+                            <PropertyValue Property="Name" String="startDateTime" />
```
baywet, line 499: Should we set the type as well?

andrueastman: I also thought about this as well, but when I was digging through this, what I found was that a CustomParameter is always a string. Validation can be enforced but there is no type of specification. 

- https://github.com/microsoft/OpenAPI.NET.OData/blob/f13fd9a453849128df87bba0e5c2ce8cb46a756e/src/Microsoft.OpenApi.OData.Reader/Vocabulary/Capabilities/CustomParameter.cs#L18
- https://github.com/oasis-tcs/odata-vocabularies/blob/main/vocabularies/Org.OData.Capabilities.V1.md#customparameter

baywet (approved):

Issue #420, opened 2023-08-30
DeviceAppManagement metadata incomplete


svrooij:
There seem the be a lot of relations missing under [App management](https://learn.microsoft.com/en-us/graph/api/resources/intune-app-conceptual?view=graph-rest-beta)

## MobileApps derived types (metadata):

https://github.com/microsoftgraph/msgraph-metadata/blob/86bdaeabf017d1c9f2dd56bd3358a67ad5675185/transforms/csdl/preprocess_csdl.xsl?wt.mc_id=SEC-MVP-5004985#L367-L377

## MobileApps derived types in the docs (missing the endpoints and relations):

- androidLobApp
- win32LobApp

Incorrect metadata means that the [Get win32LobApp](https://learn.microsoft.com/en-us/graph/api/intune-apps-win32lobapp-get?view=graph-rest-beta&wt.mc_id=SEC-MVP-5004985#http-request) is incorrect. To get Win32LobApps, you'll need at least a special filter like: `$filter=isof('microsoft.graph.win32LobApp')`

And [creating a Win32LobApp](https://learn.microsoft.com/en-us/graph/api/intune-apps-win32lobapp-create?view=graph-rest-beta&wt.mc_id=SEC-MVP-5004985) does require all the contentFiles and the committing, which is undocumented at the moment.

In [this file](https://github.com/svrooij/WingetIntune/blob/main/src/WingetIntune/Graph/GraphServiceClientExtensions.cs?wt.mc_id=SEC-MVP-5004985) you can see a lot of the endpoints for `win32LobApps` but this list is far from complete I guess.

## MobileApp add category

In the documentation is also no mention of how to add a category to an app, which can be done by posting to `deviceAppManagement/mobileApps/{id}/categories/$ref` like done [here](https://github.com/svrooij/WingetIntune/blob/21e3133b7f0b943661dc671ed22f4be9093e0b86/src/WingetIntune/Graph/GraphServiceClientExtensions.cs?wt.mc_id=SEC-MVP-5004985#L114-L132)

Issue is related to https://github.com/microsoftgraph/msgraph-sdk-dotnet/issues/2073
@andrueastman I hope I did not miss anything

Unrelated: as an external observer, is there any documentation on the metadata process? Do you start with a pre-generated version based on code and then add additional stuff using xml transformations?
Can we contribute to this as well? How does the build process work? maybe @baywet can elaborate on this topic

baywet:
[...]

PR #423, opened 2023-08-30
fix: derivedTypedFor mobileApps

svrooij:
Fixed: #420

### DIFF
--- a/transforms/csdl/preprocess_csdl.xsl
+++ b/transforms/csdl/preprocess_csdl.xsl
@@ -369,8 +369,27 @@
             <xsl:copy-of select="@* | node()" />
             <Annotation Term="Org.OData.Validation.V1.DerivedTypeConstraint">
                 <Collection>
-                    <String>microsoft.graph.mobileLobApp</String>
+                    <String>microsoft.graph.androidForWorkApp</String>
+                    <String>microsoft.graph.androidLobApp</String>
+                    <String>microsoft.graph.androidManagedStoreApp</String>
+                    <String>microsoft.graph.androidStoreApp</String>
+                    <String>microsoft.graph.iosLobApp</String>
+                    <String>microsoft.graph.iosStoreApp</String>
+                    <String>microsoft.graph.iosVppApp</String>
+                    <String>microsoft.graph.macOSDmgApp</String>
+                    <String>microsoft.graph.macOSLobApp</String>
+                    <String>microsoft.graph.macOSPkgApp</String>
+                    <String>microsoft.graph.managedAndroidLobApp</String>
+                    <String>microsoft.graph.managedIOSLobApp</String>
                     <String>microsoft.graph.managedMobileLobApp</String>
+                    <String>microsoft.graph.microsoftStoreForBusinessApp</String>
+                    <String>microsoft.graph.win32LobApp</String>
+                    <String>microsoft.graph.windowsAppX</String>
+                    <String>microsoft.graph.windowsMobileMSI</String>
+                    <String>microsoft.graph.windowsStoreApp</String>
+                    <String>microsoft.graph.windowsUniversalAppX</String>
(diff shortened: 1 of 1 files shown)


baywet (approved):
Thanks for the contribution! @irvinesunday for final review and merge

svrooij:
@baywet yeah why not, here is the PR. I've looked for all the none abstract types that have a BaseType of `mobileApp` or `mobileLobApp` and this is the complete list.

I must say I'm flying pretty blind here, but is seems to transform the beta metadata correctly

baywet:
That sounds about right to me, and I believe the reason why we were missing all those types is because when we added this transformation, they didn't exist. What might be missing are derived types from derived types of mobileApp (if any exist).
And yes, the level of complexity is why we usually don't expect external contributions here.

svrooij:
@baywet or @irvinesunday what needs to happen to move this pr forward? Would really like to see this shortcoming in the docs and sdks fixed.

I seem to not be allowed to extend the Kiota request builders, "everything is done through metadata and code generation", but this part is missing in the metadata.

baywet:
LGTM to me, waiting on @irvinesunday to review and merge

irvinesunday (approved):
LGTM

svrooij:
The documentation on the mobileContentType still seems incorrect, I'm guessing this process will take some time?

Is there "nightly" documentation available somewhere, maybe only for people with an account in the Microsoft tenant?

https://learn.microsoft.com/en-us/graph/api/intune-apps-mobileappcontent-create?view=graph-rest-beta&wt.mc_id=SEC-MVP-5004985

andrueastman:
@svrooij What specifically is seems incorrect with the docs? 
Ideally, for documentation issues, you can submit a change/suggestion by clicking the edit button on the page as below. 
image

Alternatively, the docs are open source on you can also create an issue in this repo. (The previous suggestion will also lead to creating a PR(editing a file) on the same repo.) The issue/PR should be assigned to the relevant API owner writing the doc topic.

https://github.com/microsoftgraph/microsoft-graph-docs-contrib

svrooij:
@andrueastman I'm facing a brick wall here....

Initial issue: I want to upload win32LobApps with the graph sdk for dotnet. I describe my process [here](https://svrooij.io/2023/08/31/publish-apps-to-intune/)

That means, creating the mobileAppContent, creating the mobileAppContentFile and eventually committing the file. All of which described in my post, is not available in the sdk.

Through using the portal I figured out that I need to change the type in the url.

I tried extending the request builders with extension methods, but they are not extendable [by design](https://github.com/microsoft/kiota-abstractions-dotnet/issues/113), which is a bummer...

Eventually @baywet pointed me to the metadata.... https://github.com/microsoftgraph/msgraph-sdk-dotnet/issues/2073

I have no clue how to get these endpoints in the graph sdk, please help!

svrooij:
@andrueastman to be more precise, I needed these extensions for my app (and doing it this way defeats the purpose of using the sdk in my opinion.) Can you help out to get these in the SDK/documentation? Or maybe @irvinesunday ?

https://github.com/svrooij/WingetIntune/blob/30b5f76fbaa043de94be29c4c05b692b19540b0e/src/WingetIntune/Graph/GraphServiceClientExtensions.cs

and 

https://github.com/svrooij/WingetIntune/blob/30b5f76fbaa043de94be29c4c05b692b19540b0e/src/WingetIntune/Graph/MobileAppsRequestBuilderExtensions.cs

andrueastman:
Thanks for clarifying @svrooij. 

I believe the conversion settings that generate the openApi description may not expand paths further for navigation properties of a derived type once a odata cast paths is encountered.  

Any chance you can create a separate issue for the `/contentVersions` path and then assign it to @irvinesunday or me so this may be investigated/followed up further?

Issue #434, opened 2023-09-20
Support for derivedTypes (...mobileApps/{id}/microsoft.graph.win32LobApp/...)


svrooij:
I believe the conversion settings that generate the openApi description may not expand paths further for navigation properties of a derived type once a odata cast paths is encountered.  

Any chance you can create a separate issue for the `/contentVersions` path and then assign it to @irvinesunday or me so this may be investigated/followed up further?

_Originally posted by @andrueastman in https://github.com/microsoftgraph/msgraph-metadata/issues/423#issuecomment-1727280918_

-----

Initial issue: I want to upload win32LobApps with the graph sdk for dotnet. I describe my process [here](https://svrooij.io/2023/08/31/publish-apps-to-intune/)

That means, creating the mobileAppContent, creating the mobileAppContentFile and eventually committing the file. All of which described in my post, is not available in the sdk.

Through using the portal I figured out that I need to change the type in the url.

I tried extending the request builders with extension methods, but they are not extendable https://github.com/microsoft/kiota-abstractions-dotnet/issues/113, which is a bummer...

Eventually @baywet pointed me to the metadata.... https://github.com/microsoftgraph/msgraph-sdk-dotnet/issues/2073

I have no clue how to get these endpoints in the graph sdk, please help!

I needed these extensions for my app (and doing it this way defeats the purpose of using the sdk in my opinion.)

https://github.com/svrooij/WingetIntune/blob/30b5f76fbaa043de94be29c4c05b692b19540b0e/src/WingetIntune/Graph/GraphServiceClientExtensions.cs

and

https://github.com/svrooij/WingetIntune/blob/30b5f76fbaa043de94be29c4c05b692b19540b0e/src/WingetIntune/Graph/MobileAppsRequestBuilderExtensions.cs

svrooij:
@andrueastman I cannot assign anyone, but here is the new issue.

irvinesunday:
@baywet 
Changing this setting to `true` for the Kiota-based conversion setting should suffice.
![image](https://github.com/microsoftgraph/msgraph-metadata/assets/40403681/0bedbde3-a172-4fdc-936a-e5af7dbc0470)

We will generate the expected contained nav. props defined on the target types of the type cast segments. See the below example path:

![image](https://github.com/microsoftgraph/msgraph-metadata/assets/40403681/603a7cd2-65b0-4122-bf16-21856796a395)

[...]

PR #442, opened 2023-09-22
Remove containment for `deletedItems` navigation property

irvinesunday:
Fixes https://github.com/microsoftgraph/msgraph-metadata/issues/434

Removes containment for `deletedItems` navigation property. This will help stop expanding into the type cast segments of its indexer.

### DIFF
--- a/transforms/csdl/preprocess_csdl.xsl
+++ b/transforms/csdl/preprocess_csdl.xsl
@@ -332,9 +332,11 @@
             </Annotation>
         </xsl:copy>
     </xsl:template>
+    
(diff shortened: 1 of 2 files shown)


andrueastman (approved):
👍🏼

baywet (approved):

baywet:
We'll also need to switch the setting for the conversion, I'm guessing you're going to be doing that in a separate PR @irvinesunday ?

irvinesunday:
> We'll also need to switch the setting for the conversion, I'm guessing you're going to be doing that in a separate PR @irvinesunday ?

That's correct!

Issue #438, opened 2023-09-21
`POST` missing from `https://graph.microsoft.com/v1.0/external/connections/{externalConnection-id}/schema`


sebastienlevert:
The POST operation for registering a schema on an external connection is missing from our OpenAPI description.

https://learn.microsoft.com/en-us/graph/api/externalconnectors-schema-create?view=graph-rest-1.0&tabs=http

andrueastman:
Closed via https://github.com/microsoftgraph/msgraph-metadata/pull/448

darrelmiller:
POST is not valid for this API. Create should be performed with a PATCH. We need to update the documentation once we have identified that PATCH actually works.

sebastienlevert:
@waldekmastykarz, can you confirm this worked for you?

MartinM85:
I can confirm that a schema can be created with the `PATCH`.

    PATCH /v1.0/external/connections/{connection-id}/schema

returns `202` with the `Location` header.

Now, the schema can be read

    GET /v1.0/external/connections/{connection-id}/schema

even if `GET /v1.0/external/connections/{connection-id}/operations/{op-id}` returns the status `inprogress`.

Once the schema is created, update of the schema with `PATCH` works.

The `DELETE` returns `404` and `UnknownError`

andrueastman:
Following up on the documentation fix via https://github.com/microsoftgraph/microsoft-graph-docs-contrib/issues/8672

andrueastman:
Documentation resolved via https://github.com/microsoftgraph/microsoft-graph-docs/pull/22752 and metadata https://github.com/microsoftgraph/msgraph-metadata/pull/452

PR #448, opened 2023-10-03
ExternalConnection schema

MartinM85:
Add `POST` method for `/external/connections/{externalConnection-id}/schema`
Remove `DELETE` method for `/external/connections/{externalConnection-id}/schema`

Fix for #438

I believe that schema cannot be deleted once it has been registered.

The `PATCH` method is only documented for [beta](https://learn.microsoft.com/en-us/graph/api/resources/externalconnectors-schema?view=graph-rest-beta), but should be supported for [v1.0](https://learn.microsoft.com/en-us/graph/api/resources/externalconnectors-schema?view=graph-rest-1.0) event though it is not documented.

### DIFF
--- a/transforms/csdl/preprocess_csdl.xsl
+++ b/transforms/csdl/preprocess_csdl.xsl
@@ -1314,6 +1314,21 @@
                 </xsl:when>
             </xsl:choose>
             
+            <!-- Add Insertability and remove Deletability for enternalConnection/schema navigation property-->
+            <xsl:choose>
+                <xsl:when test="not(edm:Annotations[@Target='microsoft.graph.externalConnectors.externalConnection/schema'])">
+                    <xsl:element name="Annotations">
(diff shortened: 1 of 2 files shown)


andrueastman (approved):
👍🏼

darrelmiller:
@andrueastman This will not work because you can't apply insertrestictions to a single valued navigation property.  To my knowledge doing a POST on a single valued navigation property is not a valid OData interaction and the API should never have been implemented that way.  The PATCH method should be used to set the schema.

MartinM85:
Confirmed in #438 that the PATCH can be used to set the schema. 

The DELETE returns 404.

I can fix it and keep only DeleteRestrictions.

andrueastman:
Thanks for the insights here @darrelmiller 

>I can fix it and keep only DeleteRestrictions.

@MartinM85 We're happy to let you do this

PR #452, opened 2023-10-05
External connection schema

MartinM85:
Revert added InsertRestrictions


andrueastman (approved):
Thanks again @MartinM85

Issue #480, opened 2023-11-03
Mail Message Delta list missing change type query parameter


andrueastman:
Messages delta is missing `changeType` query parameter in openAPI transform as documented at https://learn.microsoft.com/en-us/graph/api/message-delta?view=graph-rest-1.0&tabs=http#query-parameters

```http
GET /me/mailFolders/{id}/messages/delta?changeType=created
GET /users/{id}/mailfolders/{id}/messages/delta?changeType=created

GET /me/mailFolders/{id}/messages/delta?changeType=updated
GET /users/{id}/mailFolders/{id}/messages/delta?changeType=updated

GET /me/mailFolders/{id}/messages/delta?changeType=deleted
GET /users/{id}/mailFolders/{id}/messages/delta?changeType=deleted
```

PR #500, opened 2023-11-15
Adds `changeType` custom query option to delta function for `messages`

irvinesunday:
Fixes https://github.com/microsoftgraph/msgraph-metadata/issues/480

### DIFF
--- a/transforms/csdl/preprocess_csdl.xsl
+++ b/transforms/csdl/preprocess_csdl.xsl
@@ -669,7 +669,7 @@
         </xsl:copy>
     </xsl:template>
     
-    <!--Delta function for messages need the changeType parameter-->
+    <!--Delta function for messages needs the changeType parameter-->
     <xsl:template match="edm:Schema[@Namespace='microsoft.graph']/edm:Function[@Name='delta'][edm:Parameter[@Name='bindingparameter']][edm:Parameter[@Type='Collection(graph.message)']]">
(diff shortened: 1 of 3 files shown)


baywet (approved):
Thanks for the quick trunaround

transforms/csdl/preprocess_csdl.xsl
```diff
@@ -668,6 +668,26 @@
...
+        <xsl:copy>
+            <xsl:apply-templates select="@* | node()"/>
+            <Annotation Term="Org.OData.Capabilities.V1.OperationRestrictions">
+            <Record>
+                <PropertyValue Property="CustomQueryOptions">
+                    <Collection>
+                        <Record>
+                            <PropertyValue Property="Name" String="changeType" />
+                            <PropertyValue Property="Description" String="A custom query option to filter the delta response based on the type of change. Supported values are created, updated or deleted." />
+                            <PropertyValue Property="Required" Bool="true" />
```
andrueastman, line 683: @irvinesunday I believe the "required" property here should be false as its not mandatory to pass this query parameter.

irvinesunday: I must have misinterpreted this [statement](https://learn.microsoft.com/en-us/graph/api/message-delta?view=graph-rest-1.0&tabs=http#query-parameters):

image

I will push a PR to resolve it.

PR #559, opened 2024-01-19
- removes extraneous metadata files caused by the validation script during generation pipeline

baywet:


andrueastman (commented):
We'll also need to update the CI script to point to the schemas folder as removing these breaks it at

https://github.com/microsoftgraph/msgraph-metadata/blob/aa1a2b570b3e5ba346be0b53e7a0e9f300073caa/scripts/run-metadata-validation.ps1#L34C2-L34C11

baywet:
Good callout, I hadn't seen the CI failing. I updated the script to use the livesite metadata so everything points to the same thing across the board

andrueastman (approved):
Thanks for cleaning this up @baywet

PR #561, opened 2024-01-23
Set `IsComposable` to false for all functions with a `ReturnType` of `workbookRange`

irvinesunday:
Related to: https://github.com/microsoftgraph/MSGraph-SDK-Code-Generator/pull/1141

### DIFF
--- a/transforms/csdl/preprocess_csdl.xsl
+++ b/transforms/csdl/preprocess_csdl.xsl
@@ -638,6 +638,12 @@
        <xsl:attribute name="Type">Edm.Stream</xsl:attribute>
     </xsl:template>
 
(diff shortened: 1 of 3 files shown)


andrueastman (approved):

baywet:
@andrueastman @irvinesunday joining the party late here. I just wanted to make sure this change is not going to result in us pulling existing APIs away from the generated SDKs?

andrueastman:
@baywet As the conversion library(Hidi) did not support the composable functions paths until now, the paths dropped by this change on functions returning `workbookRange` had not been published before. So, the result for SDKs should be more paths exposed without the paths that would have been added for `workbookRange`. Hopefully this makes sense.

baywet:
yep, thanks for the additional context, that's all I needed.

Issue #582, opened 2024-03-04
functions range bound to workbookRangeview, workbookNamedItem, etc...


baywet:
And :
- workbookTable
- workbookTableColumn
- workbookTableRow
- workbookWorksheet

Should have `IsComposable="true"` so additional actions/functions like insert, clear, delete can be used.
This is also likely the root cause of:
- #494
- #488
- #487
- #522
- #524
- #385
- #384

andrueastman:
Reopening after https://github.com/microsoftgraph/msgraph-metadata/pull/590

MartinM85:
Hi what's the state of this ticket?

Issue #184, opened 2022-07-05
evaluate enabling git lfs on this repo


baywet:
starting a discussion here as I think there's a warning from the weekly generation logs we've probably missed over time as the OpenAPI descriptions grew.

```
remote: warning: See http://git.io/iEPt8g for more information.
remote: warning: File openapi/beta/openapi.yaml is 50.57 MB; this is larger than GitHub's recommended maximum file size of 50.00 MB
remote: warning: GH001: Large files detected. You may want to try Git Large File Storage - https://git-lfs.github.com.
```

(when pushing OpenAPI descriptions)

Currently the metadata clone of the repo takes about 8 seconds on the weekly generation process. This is not a big deal big any measure. However I suspect that clone/fetch operations are much slower on slower connections. I'm personally alone on a dedicated fiber connection, but I recognize this is not the case for everyone.

[some early LFS users report a 10X improvement](https://blog.developer.atlassian.com/git-lfs-12-clone-faster/) (note: no need to use a separate command anymore, it's just the regular git clone command).

I don't think we should optimize something if it's not a problem, but I wanted to double check whether it is a problem or not for people who often work on this repo?

(interestingly native GitHub features like contributors or commits under insight never return, and I suspect it's related)

CC @zengin @peombwa @MIchaelMainer @irvinesunday

zengin:
Thanks for raising this. I haven't observed any significant slowdown. That said, because it won't require a separate command, it shouldn't be a huge inconvenience to switch over. My vote would be on switching as the file size will only grow from this point on.

baywet:
Update: we've run into storage space issues on the agents lately this repo is 1GB once cloned, and any given agent can hold multiple copies of it because it gets recycled.
The docs repo is 1GB as well, but doesn't seem to have large files, only looots of commits.

baywet:
Additionally we might want to rewrite history with a [tool](https://github.com/rtyley/bfg-repo-cleaner/releases/tag/v1.12.5) if we don't care about the sha1's to achieve maximum gains

baywet:
re-opening this one since it wasn't implemented yet. It's just that github caught phraseology that it wasn't meant to be the case.

PR #585, opened 2024-03-05
Sets `IsComposable=true` for `workbook*` bound types

irvinesunday:
Fixes https://github.com/microsoftgraph/msgraph-metadata/issues/582

- There are no tests for this PR because to see the output the default variables values: `remove-capability-annotations` and `add-innererror-description` in the XSL file will have to be changed to `False` and `True` respectively.
- There seems to be a typo in the naming of `bindparameter` instead of `bindingparameter` in the CSDL for the respective APIs been updated. These affects both actions and functions of these APIs. The mistyped `bindparameter` is used in the XSL script so that a match is possible. When these have been resolved upstream, we shall make the necessary changes here as well. @marabooy are you able to help surface this to the relevant API owner(s)?
- We are specifically setting `IsComposable=true` for Kiota-based OpenApi generation. This is because the aftermath of this is an increase in the number of paths, which may potentially affect the PowerShell SDK. See below diffs for the expected change in the number of paths after this change goes out:
  - Powershell vs Kiota OpenApi diff.: https://www.diffchecker.com/0x24PNc6/
  - Kiota (current) vs Kiota (new) OpenApi diff: https://www.diffchecker.com/wAi8LjXR/

### DIFF
--- a/transforms/csdl/preprocess_csdl.xsl
+++ b/transforms/csdl/preprocess_csdl.xsl
@@ -644,6 +644,31 @@
         <xsl:attribute name="IsComposable">false</xsl:attribute>
     </xsl:template>
 
+    <!-- Set IsComposable to true for all functions with a bindparameter (NOT bindingparameter: possible CSDL error) of:
+         workbookRange/workbookNamedItem/workbookTable/workbookTableColumn/workbookTableRow/workbookWorksheet -->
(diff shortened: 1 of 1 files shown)


baywet (approved):

andrueastman:
@baywet @irvinesunday 

I think we may have to revert this PR temporarily to deal with some infrastructure issues. The generation pipeline isn't able to push some yamls because of the size exceeding 100MB so we would need to resolve https://github.com/microsoftgraph/msgraph-metadata/issues/184 first to enable git LFS if we are to go forward. 

Lemme know your thoughts here.

PR #590, opened 2024-03-13
Revert "Sets `IsComposable=true` for `workbook*` bound types"

andrueastman:
Temporarily Reverts microsoftgraph/msgraph-metadata#585

The generation pipeline isn't able to push some yamls because of the size exceeding 100MB so we would need to resolve https://github.com/microsoftgraph/msgraph-metadata/issues/184 first to enable git LFS if we are to go forward. 
https://github.com/microsoftgraph/msgraph-metadata/pull/585#issuecomment-1983469978

cc @baywet @irvinesunday

### DIFF
--- a/transforms/csdl/preprocess_csdl.xsl
+++ b/transforms/csdl/preprocess_csdl.xsl
@@ -644,31 +644,6 @@
         <xsl:attribute name="IsComposable">false</xsl:attribute>
     </xsl:template>
 
-    <!-- Set IsComposable to true for all functions with a bindparameter (NOT bindingparameter: possible CSDL error) of:
-         workbookRange/workbookNamedItem/workbookTable/workbookTableColumn/workbookTableRow/workbookWorksheet -->
-    <xsl:template match="edm:Schema[@Namespace='microsoft.graph']/edm:Function[edm:Parameter[@Name='bindparameter'][@Type='graph.workbookRange']] |
-                         edm:Schema[@Namespace='microsoft.graph']/edm:Function[edm:Parameter[@Name='bindparameter'][@Type='graph.workbookNamedItem']] |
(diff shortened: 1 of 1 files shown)


baywet (commented):
If I'm not mistaken, the before size is around 70MB. This would a 50% + increase. The impact on the SDKs is probably around the same.
Before we revert the change entirely, do we have an idea of what's causing the over expansion?
Customers have been requesting support for Excel endpoints for the longest time, and we're very close to enabling them.
If we have time to do so, I'd prefer waiting to merge that PR and looking into the details.
Maybe there's another contains target is composable we can switch to solve that.
But let's make sure we time box the investigation.

sebastienlevert:
The size of the resulting SDKs worries me. It's not necessarily related to this change, but if it's something that explodes our SDK size but provide very niche value, I'm proposing we think deeper about our generation logic and identify creative ways to deliver the value in some alternative ways. This might be a GREAT case of self-serve docs?

irvinesunday:
The original doc has a size of appr. 28 MBs, whereas the new one has an appr. size of 238 MBs.
Here's a diff of the paths: https://www.diffchecker.com/FmhW4g35/

The `drives` API has had an over 300% increase as a result.

andrueastman:
@baywet I think we should probably think about going ahead with reverting this PR for now given the size increases. 

As an aside we can probably look into the [NavigationType](https://github.com/oasis-tcs/odata-vocabularies/blob/main/vocabularies/Org.OData.Capabilities.V1.xml#L305:~:text=%3CEnumType%20Name=%22-,NavigationType,-%22) annotation which may not necessarily apply to functions/actions.
An alternative is also adding a conversion setting in the library to limit the expansion of paths based on depth and use that to verify the impact (significant paths added look like paths segments get repeated again once a subpath returns the same type as the base)

PR #689, opened 2024-09-13
chore: removes old powershell description

baywet:
since powershell generation now relies on PowerShell v2 description https://github.com/microsoftgraph/msgraph-sdk-powershell/blob/609948f78fbceff038e55dd09e6a2016aa0af18a/tools/DownloadOpenApiDoc.ps1#L17 we don't need these files anymore. Saving future storage and time during generation pipelines


andrueastman (approved):
👍🏼

Issue #694, opened 2024-09-23
Add XSLT to remove `autoReconcileProxyConflict` parameter from `restore` action of `graph.directoryObject`


Ndiritu:
Graph SDK releases are currently blocked because of a breaking change in the SDK generation related to introduction of a new
optional parameter `autoReconcileProxyConflict` to the `restore` action as shown below:

```xml
<Action Name="restore" IsBound="true">
    <Parameter Name="bindingParameter" Type="graph.directoryObject" Nullable="false" />
    <Parameter Name="autoReconcileProxyConflict" Type="Edm.Boolean">
        <Annotation Term="Org.OData.Core.V1.OptionalParameter" />
    </Parameter>
    <ReturnType Type="graph.directoryObject" />
    <Annotation Term="Org.OData.Core.V1.RequiresExplicitBinding" xmlns:edm="http://docs.oasis-open.org/odata/ns/edm" />
</Action>
```

This [issue](https://github.com/microsoft/OpenAPI.NET.OData/issues/582) has been created to unblock the SDK generation issue and would require further work on Kiota after this issue is fixed.

To unblock weekly SDK releases, could we temporarily remove this parameter?

Further Ref:
- [Workload change](https://dev.azure.com/msazure/One/_git/AD-AggregatorService-Workloads/commit/693d356f0ca43800656c39ddf87e0a460beed382?path=/Workloads/Microsoft.DirectoryServices/override/schema-Prod-v1.0.csdl&version=GBmaster&_a=files)
- [SDK generation discussion](https://github.com/microsoftgraph/msgraph-sdk-java/pull/2163)

PR #697, opened 2024-09-25
Remove autoReconcileProxyConflict parameter from restore action of graph.directoryObject

andrueastman:
Fixes https://github.com/microsoftgraph/msgraph-metadata/issues/694

Temporary fix for metadata as https://github.com/microsoft/OpenAPI.NET.OData/issues/582 is being resolved


Ndiritu (approved):

baywet (approved):

zaaj:
Does this PR break the API's ability to auto-Reconcile Proxy conflicts at restore-time? This means that if there's a critical account that got soft-deleted, and another account took over one of it's proxyAddresses, we cannot restore the soft-deleted account, is that correct?

PR #742, opened 2024-12-27
Schema sync 2024-12-27 from 1.3.531.0

marabooy:
Schema sync 2024-12-27 from 1.3.531.0


Ndiritu:
same version has successful build https://github.com/microsoftgraph/msgraph-metadata/pull/740

PR #765, opened 2025-02-06
Adds update restrictions to update sitePage properties

### LINKED ISSUE #764: Add UpdateMethod PATCH in UpdateRestrictions annotations for sitePage
Based on the Api reference document here https://learn.microsoft.com/en-us/graph/api/sitepage-update?view=graph-rest-1.0&tabs=http the open Api file should have a `Patch` operation

timayabi2020:
Fixes #764

### DIFF
--- a/transforms/csdl/preprocess_csdl.xsl
+++ b/transforms/csdl/preprocess_csdl.xsl
@@ -991,6 +991,15 @@
             </xsl:call-template>
         </xsl:copy>
     </xsl:template>
+    <!-- Add paths for user authentication requirements by adding annotations to update properties of the sitePage object-->
+    <xsl:template match="edm:Schema[@Namespace='microsoft.graph']/edm:EntityType[@Name='sitePage']">
+        <xsl:copy>
+            <xsl:apply-templates select="@* | node()"/>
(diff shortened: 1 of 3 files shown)


irvinesunday:
> If you take a look at the openApi files generated for this PR at https://github.com/microsoftgraph/msgraph-metadata/actions/runs/13180054025?pr=765
> 
> The patch method is generated for `'/sites/{site-id}/pages/{baseSitePage-id}'` but not for the `'/sites/{site-id}/pages/{baseSitePage-id}/graph.sitePage'` which doesn't look to be what we want here. There is the possibility the conversion library doesn't support the annotation for odata cast segments but maybe @irvinesunday could confirm.

The conversion lib supports only `Get` for OData type cast segments: https://github.com/microsoft/OpenAPI.NET.OData/blob/efaf60a8c187c4458c31d19b3ea5b88664ec56e7/src/Microsoft.OpenApi.OData.Reader/PathItem/ODataTypeCastPathItemHandler.cs#L22-L25
image

timayabi2020:
@irvinesunday thanks for confirmation. So how do I add the patch operation for that path (`/sites/{site-id}/pages/{page-id}/microsoft.graph.sitePage`)

PR #811, opened 2025-04-15
Schema sync 2025-04-15 from 1.3.719.0

marabooy:
Schema sync 2025-04-15 from 1.3.719.0


musale (approved):

baywet:
After some more spelunking, this was only recently introduced with #801. So a "regression" is expected here.

PR #848, opened 2025-07-11
Schema sync 2025-07-11 from 1.4.47.0

marabooy:
Schema sync 2025-07-11 from 1.4.47.0


adrian05-ms:
Close due to no changes

PR #923, opened 2025-10-28
fix: set /Places insertable = true

MIchaelMainer:
API has new capability. We had disabled the capability for generation when it wasn't available. Now that it is, we need to re-enable.


transforms/csdl/preprocess_csdl.xsl
```diff
@@ -1748,7 +1748,7 @@
                                 <xsl:with-param name="readable">false</xsl:with-param>
                             </xsl:call-template>
                             <xsl:call-template name="InsertRestrictionsTemplate">
-                                <xsl:with-param name="insertable">false</xsl:with-param>
+                                <xsl:with-param name="insertable">true</xsl:with-param>
```
gavinbarron: If we can remove the node here completely IMO that would be better to keep us closer to what was originally produced.

PR #1094, opened 2026-05-04
Ramsess/add casing check to metadata

ramsessanchez:
Adds a check to our schema syncs so casing changes in entityTypes, complexTypes, and enumTypes are flagged. The following is achieved by:
- Add script to create json files containing maps for entityTypes, complexTypes, and enumTypes for both v1.0 and beta. 
- Add a second script to compare the casing of type names in the schema csdl files to what is stored in the json maps. 
- An error is thrown if a type exists in the json map and the csdl with different casing; new types and deleted types are not flagged.
- Adds a github action that runs on pull requests making changes to the schema csdl files. This would target the pr's created by our generation pipeline. Will flag changes to casing changes as breaking. 
- If a pr is approved and merged then the action runs the mapping script again to set a new baseline for what is contained in our csdl files. 

Source of issue described here: https://github.com/microsoft/kiota/issues/7654?reload=1?reload=1

### DIFF
--- a/.github/workflows/check-casing-changes.yml
+++ b/.github/workflows/check-casing-changes.yml
@@ -47,6 +47,8 @@ jobs:
   update-baselines:
     if: github.event_name == 'push'
     runs-on: ubuntu-latest
+    permissions:
+      contents: write
 
     steps:
     - uses: actions/checkout@v6
--- a/.github/workflows/check-casing-changes.yml
+++ b/.github/workflows/check-casing-changes.yml
@@ -1,0 +1,78 @@
+# Validates that PRs don't introduce unintentional casing changes to type/enum names,
+# members, or properties in CSDL schema files. After a PR merges, automatically
+# regenerates the type-mapping baseline files to keep them current.
+
+name: Check type and enum casing changes
+
+on:
+  push:
+    branches: [ master ]
+    paths:
+      - 'schemas/*-Prod.csdl'
+  pull_request:
+    branches: [ master ]
+    paths:
+      - 'schemas/*-Prod.csdl'
+
+jobs:
+  # Runs on PRs: compares modified CSDL files against the committed type-mapping
+  # JSON baselines. Fails if any name matches case-insensitively but differs
+  # case-sensitively (e.g., "accessLevel" -> "AccessLevel"). New types are allowed.
+  check-casing:
+    if: github.event_name == 'pull_request'
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v6
+      with:
+        fetch-depth: 0
+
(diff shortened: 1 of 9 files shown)


.github/workflows/check-casing-changes.yml
```diff
@@ -0,0 +1,78 @@
...
+      shell: pwsh
+      run: |
+        $files = "${{ steps.changed-files.outputs.files }}" -split ','
+        ./scripts/check-casing-changes.ps1 -CsdlFiles $files
+
+  # Runs on push to master (after PR merge): regenerates the type-mapping baseline
+  # JSON files from the updated CSDL schemas and auto-commits them if changed.
+  # This keeps baselines current so future PRs are checked against the latest state.
+  update-baselines:
+    if: github.event_name == 'push'
```
MIchaelMainer, line 48: Add explicit perm grant

```
    permissions:
      contents: write
```

scripts/check-casing-changes.ps1
```diff
@@ -0,0 +1,272 @@
...
+    <#
+    .Synopsis
+        Checks ComplexType or EntityType casing against baseline.
+    #>
+    param(
+        [System.Xml.XmlDocument]$Csdl,
+        [System.Xml.XmlNamespaceManager]$NsMgr,
+        [PSCustomObject]$Baseline,
+        [string]$TypeKind,
+        [string]$Version
```
MIchaelMainer, line 133: $Version is not used.

scripts/check-casing-changes.ps1
```diff
@@ -0,0 +1,272 @@
...
+    # Not found in baseline — new name, no violation
+    return $null
+}
+
+function Test-EnumTypeCasing {
+    param(
+        [System.Xml.XmlDocument]$Csdl,
+        [System.Xml.XmlNamespaceManager]$NsMgr,
+        [PSCustomObject]$Baseline,
+        [string]$Version
```
MIchaelMainer, line 64: $Version is not used.

MIchaelMainer (changes requested):

.github/workflows/check-casing-changes.yml
ramsessanchez: added

scripts/check-casing-changes.ps1
ramsessanchez: version is now used in the failure output to clearly state the version and namespace the failure happened in.

scripts/check-casing-changes.ps1
ramsessanchez: using it for output now

MIchaelMainer (approved):

